PR #3, opened 2025-06-20
Create scaffolding for REST API and device-gateway server and handler logic

doanac:


vkhoroz (commented):
LGTM

This is a huge load of work. Thanks!

PR #58, opened 2025-12-05
auth: Move scopes to bitmask model

doanac:
Doing string comparisons is becoming difficult and error prone for new code being written to manage permissions and tokens. This approach makes it easier to manage and test scopes.


vkhoroz (approved):
LGTM

PR #119, opened 2026-02-18
Rate limits

doanac:


doanac:
@vkhoroz - here is a work-in-progress idea for rate-limits and blocking users. I think this may be a better approach than https://github.com/foundriesio/dg-satellite/pull/113 and solve a better problem than what the original PR was doing. 

take your time - this was weird to sort out. I was surprised how much code I had to come up with to do this; I expected to find libraries that could handle more of this.

doanac:
@vkhoroz - i've re-worked this a couple times. This current version feels like it getting closer to something we could live with. No rush - but take a look and see what you think.

auth/provider_local_ratelimter.go
```diff
@@ -0,0 +1,142 @@
...
+	if allow, err := rl.store.Allow(identifier); !allow {
+		if !exists {
+			r := float64(1) / 60
+			v = &visitor{
+				Limiter:  rate.NewLimiter(rate.Limit(r), rl.badAuthLimit),
+				lastSeen: now,
+			}
+			rl.visitors[identifier] = v
+		}
+		v.blockUntil = now.Add(rl.attemptsBlockDuration)
```
vkhoroz, line 156: IIUC this is the per-second rate-limiter block competing with the bad-auth rate-limiter for the place within the same map (see e.g. line 95).  It seems to me that you've made a good deal of tracking that competition properly.

I think it only deserves a small comment to explain it. For example, I was worried that a currently set `blockUntil` might be already set to a later time than what we're trying to set... but this is already handled by line 122 above.

auth/provider_local_ratelimter.go
```diff
@@ -0,0 +1,142 @@
...
+	v, exists := rl.visitors[identifier]
+	if !exists {
+		// Allow 5 bad auth operations per minute before blocking,
+		// rate-limiter doesn requests/second, this converts to requests/minute
+		r := float64(1) / 60
+		v = &visitor{
+			Limiter:  rate.NewLimiter(rate.Limit(r), rl.badAuthLimit),
+			lastSeen: time.Now(),
+		}
+		rl.visitors[identifier] = v
```
vkhoroz: This code is copied in two places - a good candidate to extract into smth like `v = rl.newVisitor(identifier, now)`

auth/provider_local_ratelimter.go
```diff
@@ -0,0 +1,142 @@
...
+func (rl *localAuthRateLimiter) FlagBadOperation(c echo.Context) {
+	rl.mutex.Lock()
+	defer rl.mutex.Unlock()
+
+	identifier := c.RealIP()
+	v, exists := rl.visitors[identifier]
+	if !exists {
+		// Allow 5 bad auth operations per minute before blocking,
+		// rate-limiter doesn requests/second, this converts to requests/minute
+		r := float64(1) / 60
```
vkhoroz, line 95: This is actually a `const MinutesPerSecond float64 = 1.0 / 60.0`

auth/provider_local_ratelimter.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	// Check if this IP is already blocked
+	v, exists := rl.visitors[identifier]
+	if exists && (now.Before(v.blockUntil) || v.Tokens() <= 0) {
+		return false, errTooManyBadAuthOps
+	}
+
+	// Check the per-second rate limit; if exceeded, block the IP for 1 minute
+	if allow, err := rl.store.Allow(identifier); !allow {
```
vkhoroz: Just IMHO we don't need to re-use this echo's middleware at all.
We save maybe some 3-5 lines of code, but drag a dependency which kind of copies the entire visitors logic with separate mutexes and garbage collection.

What we can do instead is declare two `rate.Limiter` fields inside our visitor class: one for bad-auth-per-minute-limit and one for auth-attempt-per-second-limit, and use their `AllowN` methods directly (like you already do inside `FlagBadOperation`. That would save a lot of computational logic, which is currently hidden behind this deceptive call to `store.Allow`.

auth/provider_local_ratelimter.go
```diff
@@ -0,0 +1,142 @@
...
+	now := time.Now()
+	if time.Since(rl.lastGc) > 5*time.Minute {
+		for id, v := range rl.visitors {
+			// we flag operations by the minute, so 2 minutes is plenty
+			if time.Since(v.lastSeen) > 2*time.Minute && now.After(v.blockUntil) {
+				delete(rl.visitors, id)
+			}
+		}
+		rl.lastGc = now
+	}
```
vkhoroz, line 129: This GC part is what I'm a bit worried about. Under load we might end up with a staggering number of requests waiting for a mutex, while the single request which hit the GC logic is trying to clean up thousands of visitors. We give an attacker an easy target to DDoS.

This is often sorted out by "generations", and in our case we actually only need two generations: front and back.
Here is how it works:
1. We declare two maps of visitors `rl.front` and `rl.back`.
2. When we look up for a visitor - we look into `rl.front` and then into `rl.back`.
3. If a visitor exists in `rl.back` (but not `rl.front`) - it is moved to `rl.front`.
4. Each new visitor is added to `rl.front`.
5. Once upon a time (e.g. twice the maximum block interval) we perform a simple pointer swap: `rl.back = rl.front; rl.front = make(map)`.

What we effectively do that way:
1. Active visitors are always moved to front and stay there (as long as they are still active).
2. Stale visitors are garbage collected instantly in huge chunks without a necessity to loop over each visitor.

This kind of garbage collection is constant regardless of the number of visitors.
A small price to pay is a marginally slower visitor lookup, which is still close to zero and almost constant.

WDYT?

auth/provider_local_ratelimter.go
```diff
@@ -0,0 +1,142 @@
...
+
+	identifier := c.RealIP()
+	v, exists := rl.visitors[identifier]
+	if !exists {
+		// Allow 5 bad auth operations per minute before blocking,
+		// rate-limiter doesn requests/second, this converts to requests/minute
+		r := float64(1) / 60
+		v = &visitor{
+			Limiter:  rate.NewLimiter(rate.Limit(r), rl.badAuthLimit),
+			lastSeen: time.Now(),
```
vkhoroz: This `lastSeen` get overwritten below (line 97).

auth/provider_local_ratelimter.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+		rl.visitors[identifier] = v
+	}
+	v.lastSeen = time.Now()
+	allow := v.AllowN(v.lastSeen, 1)
+	if !allow {
+		v.blockUntil = time.Now().Add(rl.badAuthBlockDuration)
+		context.CtxGetLog(c.Request().Context()).Warn("Too many bad auth operations. Blocking IP", "ip", identifier, "until", v.blockUntil)
```
vkhoroz: This "trivial" line may become a bottleneck, as it is a heavy I/O operation which is still covered by a mutex lock - something that must have been released as soon as possible.

Thus, I'd split this method into two parts - operating on visitors (covered by mutex) and anything else (outside the lock). Currently, this is a single line which matches the "anything else" definition, but it is worth a deal.

auth/provider_local_ratelimter.go
```diff
@@ -0,0 +1,142 @@
...
+		if !exists {
+			r := float64(1) / 60
+			v = &visitor{
+				Limiter:  rate.NewLimiter(rate.Limit(r), rl.badAuthLimit),
+				lastSeen: now,
+			}
+			rl.visitors[identifier] = v
+		}
+		v.blockUntil = now.Add(rl.attemptsBlockDuration)
+		slog.Warn("Rate limit exceeded. Blocking IP", "ip", identifier, "until", v.blockUntil, "error", err)
```
vkhoroz: I think this log can be a part of the `DenyHandler`. That way is is not covered by a mutex, and we can also access the request context over there.

auth/provider_local_ratelimter.go
```diff
@@ -0,0 +1,142 @@
...
+		return false, errTooManyBadAuthOps
```
vkhoroz: IIUC this is a plain "per-second" rate limit being hit i.e. not really a "too-many-bad-auth".

vkhoroz (commented):
I like this - it does what we need.

My comments are primarily about how to make this middleware's footprint smaller and make the time spent inside a mutex lock predictable.

auth/provider_local_ratelimter.go
vkhoroz: What might sound complex, is actually trivial in coding:
```
type visitorMap struct {
	front map[string]*visitor
	back  map[string]*visitor
}

func (m *visitorMap) get(id string) visitor {
	v, ok := m.front[id]
	if !ok {
		if v, ok = m.back[id]; ok {
			// We may actually skip this, as `m.back` is anyway garbage collected later.
			// There is no optimization from a memory perspective, as the map size never shrinks down IIRC, and we only store a pointer here.
			delete(m.back, id)
		} else {
			v = &visitor{...}
		}
		m.front[id] = v
	}
	return v
}

func (m *visitorMap) gc() {
	m.back = m.front  // garbage collects m.back
	m.front = make(map[string]*visitor, 1000)
}

func newVisitorMap() *visitorMap {
	return &visitorMap{
		make(map[string]*visitor, 1000), // pre-allocate enough space to avoid re-sizing.
		make(map[string]*visitor, 0), // initially, back generation is empty
	}
}
```

doanac:
@vkhoroz - this latest force-push is a lot closer to what you want. It doesn't have the optimization for GC logic.

However, as I'm reading this, I've decided this should go one layer lower and be available to *all* auth providers. e.g. we should do rate limits for API requests that use the wrong token.

I'm going to work on a new PR that will look a lot like this in structure, but be managed differently from a config and auth provider standpoint.

PR #132, opened 2026-04-07
Fix issues around authentication, UI->REST_API, and CSRF

doanac:
I've done a couple of things dealing with how the web ui should talk to the backend API. This set of changes:

 * reverts all the original changes
 * stops storing the web session ID in clear text in the db
 * fixes our cookie expiration logic .. again
 * Adds comprehensive CSRF protection

The CSRF protection commit is the main thing to look at. It:
 * sends a cookie with each session that has the CSRF token
 * checks during REST APIs for cookie auth and if the caller provides the CSRF token

### DIFF
--- a/auth/csrf.go
+++ b/auth/csrf.go
@@ -1,0 +1,67 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+package auth
+
+import (
+	"crypto/rand"
+	"crypto/subtle"
+	"net/http"
+	"time"
+
+	"github.com/labstack/echo/v4"
+)
+
+const CsrfCookieName = "dg-satellite-csrf"
+const CsrfHeaderName = "X-CSRF-Token"
+
+// SetCsrfCookie sets a CSRF cookie on the response. It should be called
+// when a new session is created. The cookie is NOT HttpOnly so that
+// JavaScript can read the value to include in request headers.
+func SetCsrfCookie(c echo.Context, expires time.Time) {
+	c.SetCookie(&http.Cookie{
+		Name:     CsrfCookieName,
+		Value:    rand.Text(),
+		Path:     "/",
+		Expires:  expires,
+		HttpOnly: true,
+		Secure:   true,
+		SameSite: http.SameSiteStrictMode,
+	})
+}
+
+// CsrfCheck is a middleware that validates the CSRF token for non-safe HTTP
+// methods (anything other than GET, HEAD, OPTIONS). It skips the check for
+// requests that use an Authorization header (i.e. API token auth).
+func CsrfCheck(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		method := c.Request().Method
+		if method == http.MethodGet || method == http.MethodHead || method == http.MethodOptions {
+			return next(c)
+		}
+
+		// Skip CSRF check for token-authenticated API requests
+		if c.Request().Header.Get("Authorization") != "" {
+			return next(c)
+		}
+
+		cookie, err := c.Cookie(CsrfCookieName)
+		if err != nil || cookie.Value == "" {
+			return c.String(http.StatusForbidden, "Missing CSRF cookie")
+		}
+
+		headerToken := c.Request().Header.Get(CsrfHeaderName)
+		if headerToken == "" {
+			headerToken = c.FormValue("_csrf")
+		}
+		if headerToken == "" {
+			return c.String(http.StatusForbidden, "Missing CSRF token")
+		}
+
+		if subtle.ConstantTimeCompare([]byte(cookie.Value), []byte(headerToken)) != 1 {
+			return c.String(http.StatusForbidden, "CSRF token mismatch")
+		}
+
+		return next(c)
+	}
+}
--- a/auth/provider_common.go
+++ b/auth/provider_common.go
@@ -8,7 +8,6 @@ import (
 	"log/slog"
 	"net/http"
 	"strings"
-	"time"
 
 	"github.com/foundriesio/dg-satellite/storage/users"
 	"github.com/labstack/echo/v4"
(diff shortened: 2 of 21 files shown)


auth/csrf.go
```diff
@@ -0,0 +1,30 @@
...
+// SetCsrfCookie sets a CSRF cookie on the response. It should be called
+// when a new session is created. The cookie is NOT HttpOnly so that
+// JavaScript can read the value to include in request headers.
+func SetCsrfCookie(c echo.Context, expires time.Time) {
+	c.SetCookie(&http.Cookie{
+		Name:     CsrfCookieName,
+		Value:    rand.Text(),
+		Path:     "/",
+		Expires:  expires,
+		HttpOnly: false,
```
vkhoroz: I don't quite get why is this cookie NOT HttpOnly.

CSRF is usually passed via hidden controls.
It can also be passed inside a user session (secure cookie) if we want to make our server side stateless.
And that's exactly what I see done in this PR!!!

IIUC if we set HttpOnly to false - JavaScript would be able to get a value of this cookie, and inject it into whatever request they want. A malicious JavaScript, hence, can do that too - an opposite to what we want to achieve.

server/ui/api/middleware.go
```diff
@@ -57,36 +56,3 @@ func gzipContentTypeAsContentEncoding(next echo.HandlerFunc) echo.HandlerFunc {
 		return next(c)
 	}
 }
-
-// csrfCheck is a middleware that validates the CSRF token for non-safe HTTP
-// methods (anything other than GET, HEAD, OPTIONS). It skips the check for
-// requests that use an Authorization header (i.e. API token auth).
-func csrfCheck(next echo.HandlerFunc) echo.HandlerFunc {
```
vkhoroz: Can you put this function immediately where you see it fit the best, in the previous commit?

auth/provider_oauthbase.go
```diff
@@ -121,7 +127,9 @@ func (p oauth2BaseProvider) handleOauthCallback(c echo.Context) error {
 		return err
 	}
 
+	fmt.Println("ANDY session timeout?", p.sessionTimeout)
 	expires := time.Now().Add(p.sessionTimeout)
+	fmt.Println("ANDY session expire?", expires)
 	sessionId, err := user.CreateSession(c.RealIP(), expires.Unix(), user.AllowedScopes)
```
vkhoroz, line 131: Debugging is nice :)

server/ui/server.go
```diff
@@ -46,6 +46,7 @@ func NewServer(ctx context.Context, db *storage.DbHandle, fs *storage.FsHandle,
 	daemons := daemons.New(ctx, strg, users)
 
 	srv := server.NewServer(ctx, e, serverName, port, nil)
+	e.Use(auth.CsrfCheck)
```
vkhoroz, line 49: Do we need the CSRF check in the entire UI?
I thought we only need it when the request is made to a different `host:port` from where the site is hosted.
I'm a bit lost now.

Mind me, this is just a question.

auth/provider_common.go
```diff
@@ -65,9 +64,6 @@ func (p *commonProvider) GetSession(c echo.Context) (*Session, error) {
 	} else if len(cookie.Value) == 0 {
 		return nil, p.renderer.renderLoginPage(c, "")
 	}
-	if cookie.Expires.Before(time.Now()) {
-		return nil, p.renderer.renderLoginPage(c, "Cookie expired")
-	}
```
vkhoroz, line 74: Hmm....

I thought that validating cookie expiration on the server is crucial to prevent replay attacks.
This has nothing to do with how a browser works.

An attacker can steal a cookie somehow (e.g. from the user's disk or browser).
If cookie expiration is not validated server-side, that cookie is technically valid and can be used forever, even after it was cleared from the browser.

auth/csrf.go
doanac: The cookie is same-site-strict so its not available this way.

auth/provider_common.go
doanac: A cookie is just a value you send back. The expiry for it is advisory for the browser to manage it. When a user effectively runs:
```
 curl -H "Cookie: session-id-value" https://example.com
```
There's no way for a server to know its expired - that's why you track the expiry in your backend session data.

server/ui/server.go
doanac: I found it easier and safer to apply a check globally so you don't have to think about which specific APIs to protect. What we are trying to protect against is something like this. malicious.com has a page with:
```
<html>
  <body>
    <form action="https://dg-satellite.com/v1/devices/XXX" method="DELETE">
    </form>
    <script>
      document.forms[0].submit();
    </script>
  </body>
</html>
```

We have two means of protection against this:
 * The CSRF token is same-site, so clicking this button would never include it in the request.
 * The malicious page won't know what the value of this token is that they need to include in the request.
 * It would also require they somehow had a user's session or the session cookie wasn't sufficiently safe.

server/ui/api/middleware.go
doanac: fixed in force push

auth/csrf.go
vkhoroz: There are different ways to run CSRF attacks, including site embedding (both ways), direct API calls, or even cross-browser-tab scripts. I'm not expert enough to craft an attack to break a current way of doing things. But, I know that for security we always need to give the least possible privilege, which is enough to satisfy our site's needs.

A way our web UI is currently implemented would work well enough with HttpOnly cookie, as we never access it in JavaScript. So, it's better to keep it HttpOnly as a more restrictive policy.

server/ui/server.go
vkhoroz: Sounds reasonable. At least this does not break things, so it's good to go.

auth/provider_common.go
vkhoroz: Let me expand my question...

Do we check if the cookie expired somewhere? By that, I don't mean the Expires property of a Cookie (which is indeed advisory for the browser), but a cookie expiration stored in e.g. the database or inside a cookie value.
I think we should. And I think you are right that we need not check the `cookie.expires`.

doanac: The session token has an expiration which is the same as what we set when we create  the cookie

auth/csrf.go
doanac: could call. Fixed in force push

vkhoroz (approved):
LGTM

auth/csrf.go
vkhoroz: Thanks for this.

The point is: if our site JavaScript ever needs this CSRF token - it can get it from the hidden input.
It should never need to read a cookie for that.

PR #134, opened 2026-04-10
auth: Fix redirect for oauth provider

doanac:
When moving to a SameSiteStrict cookie, we broke the redirect that happens after we recieve the oauth response from the provider. Since the cookie is "SameSiteStrict", the redirect which is coming from github/google, won't include the session cookie which was resulting in a login page. You could simply hit "refresh" at that point and see the correct page. Rather than change to use a "Lax" cookie, this change causes the browser to do a reload which will send the cookie.


vkhoroz (approved):
This is smart... and kind of explains why we often see those "redirecting..." pages on so many sites.

PR #135, opened 2026-04-14
Make swagger docs functional

doanac:
Our swagger comments needed a little cleaning up. This set of changes improves what we render for swagger docs and also adds checks to validate it during PRs as well as including the swagger docs with each release of the project

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 *.swo
 .extra-certs
 .compose-server-data
+docs/swagger/*
+bin/*
+release/*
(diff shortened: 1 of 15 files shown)


.gitignore
```diff
@@ -2,3 +2,4 @@
 *.swo
 .extra-certs
 .compose-server-data
+docs/swagger/*
```
vkhoroz, line 5: Can you add `release/*` to the `.gitignore`?

I also wanted to add `bin/*`, but somehow it is not yet here. Can you add that too?

vkhoroz (approved):
Nice work!

.gitignore
doanac: updated in force push

PR #136, opened 2026-04-14
Device pagination sorting

doanac:
Add support for pagination in CLI and WebUI

### DIFF
--- a/cli/subcommands/devices/list.go
+++ b/cli/subcommands/devices/list.go
@@ -61,7 +61,7 @@ func init() {
 	DevicesCmd.AddCommand(listCmd)
 	listCmd.Flags().StringP("columns", "", "uuid,target,last-seen",
 		"Comma-separated list of columns to display (available: "+colmnsStr+")")
-	listCmd.Flags().IntP("page", "p", 0, "Page number to display (0-indexed)")
+	listCmd.Flags().IntP("page", "p", 1, "Page number to display")
 	listCmd.Flags().StringP("sort", "s", "", "Sort order for devices ("+sortStr+")")
 }
 
--- a/cli/subcommands/devices/list.go
+++ b/cli/subcommands/devices/list.go
@@ -26,17 +26,20 @@ var allColumns = []string{
 	"labels",
 }
 
+const defaultPageLimit = 50 // the number of devices to fetch per page when listing.
+
 var listCmd = &cobra.Command{
 	Use:   "list",
-	Short: "List all devices",
-	Long:  `List all devices known to the server`,
+	Short: "List devices",
+	Long:  `List devices known to the server. By default shows the first page of results.`,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		columns, err := validateColumns(cmd.Flag("columns").Value.String())
 		if err != nil {
 			return err
 		}
+		page, _ := cmd.Flags().GetInt("page")
 		api := api.CtxGetApi(cmd.Context())
-		listDevices(api.Devices(), columns)
+		listDevices(api.Devices(), columns, page)
 		return nil
 	},
 }
@@ -46,6 +49,7 @@ func init() {
 	DevicesCmd.AddCommand(listCmd)
 	listCmd.Flags().StringP("columns", "", "uuid,target,last-seen",
 		"Comma-separated list of columns to display (available: "+colmnsStr+")")
+	listCmd.Flags().IntP("page", "p", 0, "Page number to display (0-indexed)")
 }
 
 func validateColumns(columnsStr string) ([]string, error) {
@@ -58,8 +62,8 @@ func validateColumns(columnsStr string) ([]string, error) {
 	return columns, nil
 }
 
-func listDevices(dapi api.DeviceApi, columns []string) {
-	devices, err := dapi.List()
+func listDevices(dapi api.DeviceApi, columns []string, page int) {
+	devices, hasMore, totalPages, err := dapi.ListPage(page, defaultPageLimit)
 	cobra.CheckErr(err)
 
 	headers := make([]string, 0, len(columns))
@@ -77,6 +81,9 @@ func listDevices(dapi api.DeviceApi, columns []string) {
 	}
 
 	table.Render()
+	if hasMore {
+		fmt.Printf("\nA total of %d pages of devices available. Use '--page %d' for the next page.\n", totalPages, page+1)
+	}
 }
 
 func getColumnValue(device *api.DeviceListItem, column string) string {
--- a/cli/subcommands/devices/list.go
+++ b/cli/subcommands/devices/list.go
@@ -37,33 +37,56 @@ var listCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
+		sortBy, _ := cmd.Flags().GetString("sort")
+		if err := validateSortBy(sortBy); err != nil {
+			return err
+		}
 		page, _ := cmd.Flags().GetInt("page")
 		api := api.CtxGetApi(cmd.Context())
-		listDevices(api.Devices(), columns, page)
+		listDevices(api.Devices(), columns, page, sortBy)
 		return nil
 	},
 }
 
+var validSortValues = []string{
+	"name-asc", "name-desc",
+	"created-at-asc", "created-at-desc",
+	"last-seen-asc", "last-seen-desc",
+	"uuid-asc", "uuid-desc",
+}
+
 func init() {
 	colmnsStr := strings.Join(allColumns, ",")
+	sortStr := strings.Join(validSortValues, ", ")
 	DevicesCmd.AddCommand(listCmd)
 	listCmd.Flags().StringP("columns", "", "uuid,target,last-seen",
 		"Comma-separated list of columns to display (available: "+colmnsStr+")")
 	listCmd.Flags().IntP("page", "p", 0, "Page number to display (0-indexed)")
+	listCmd.Flags().StringP("sort", "s", "", "Sort order for devices ("+sortStr+")")
+}
+
+func validateSortBy(sortBy string) error {
+	if sortBy == "" {
+		return nil
+	}
+	if !slices.Contains(validSortValues, sortBy) {
+		return fmt.Errorf("invalid sort value: %s (valid: %s)", sortBy, strings.Join(validSortValues, ", "))
+	}
+	return nil
 }
 
 func validateColumns(columnsStr string) ([]string, error) {
 	columns := strings.Split(columnsStr, ",")
 	for _, col := range columns {
-		if slices.Index(allColumns, col) < 0 {
+		if !slices.Contains(allColumns, col) {
 			return nil, fmt.Errorf("invalid column: %s", col)
 		}
 	}
 	return columns, nil
 }
 
-func listDevices(dapi api.DeviceApi, columns []string, page int) {
-	devices, hasMore, totalPages, err := dapi.ListPage(page, defaultPageLimit)
+func listDevices(dapi api.DeviceApi, columns []string, page int, sortBy string) {
+	devices, hasMore, totalPages, err := dapi.ListPage(page, defaultPageLimit, sortBy)
 	cobra.CheckErr(err)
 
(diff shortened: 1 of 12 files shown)


doanac:
This adds column sorting in the UI:
image

And pagination:
image

doanac:
I've tested this with 10,000 devices and things remain very smooth.

I also added a change to allow you to jump to the last page of results:
image

storage/api/api_storage.go
```diff
@@ -206,21 +208,27 @@ func NewStorage(db *storage.DbHandle, fs *storage.FsHandle) (*Storage, error) {
...
 	}
 	stmt, ok := s.stmtDeviceList[orderBy]
 	if !ok {
-		return nil, fmt.Errorf("invalid order by arg: %s", opts.OrderBy)
+		return nil, 0, fmt.Errorf("invalid order by arg: %s", opts.OrderBy)
+	}
+
+	total, err := s.stmtDeviceCount.run()
+	if err != nil {
+		return nil, 0, err
```
vkhoroz, line 223: I'd rather make this a separate function like e.g. `DevicesCount`.
We might have an API which does not need this count, or only needs a count.
Also, an API might be improved later to run these two functions in parallel.

fwiw, just my gut feeling.

storage/api/api_storage_test.go
```diff
@@ -113,7 +113,7 @@ func TestDeviceDelete(t *testing.T) {
 	require.Nil(t, err)
 	require.Nil(t, d, "deleted device should not be returned by DeviceGet")
 
-	devices, err := s.DevicesList(DeviceListOpts{Limit: 100})
+	devices, _, err := s.DevicesList(DeviceListOpts{Limit: 100})
```
vkhoroz: I think we want to check this count too in unit tests.

cli/subcommands/devices/list.go
```diff
@@ -26,17 +26,20 @@ var allColumns = []string{
 	"labels",
 }
 
+const defaultPageLimit = 1000 // the number of devices to fetch per page when listing.
```
vkhoroz: Hmm.... this default does not look sensible. I might easily blow off the terminal scrolling history.
Can we make it more conservative, like e.g. 100 or even 50?

cli/subcommands/devices/list.go
```diff
@@ -77,6 +81,9 @@ func listDevices(dapi api.DeviceApi, columns []string) {
 	}
 
 	table.Render()
+	if hasMore {
+		fmt.Printf("\nMore results available. Use --page %d for the next page.\n", page+1)
```
vkhoroz: I think this new API allows to easily deduce how many more pages there are in total.
That'd be a valuable information to show to the user, like e.g. `A total of 42 pages of devices available. Use '--page 17' for the next page.`.

cli/subcommands/devices/list.go
```diff
@@ -37,19 +37,42 @@ var listCmd = &cobra.Command{
...
 		"Comma-separated list of columns to display (available: "+colmnsStr+")")
 	listCmd.Flags().IntP("page", "p", 0, "Page number to display (0-indexed)")
+	listCmd.Flags().StringP("sort", "s", "", "Sort order for devices ("+sortStr+")")
+}
+
+func validateSortBy(sortBy string) error {
+	if sortBy == "" {
+		return nil
+	}
+	if slices.Index(validSortValues, sortBy) < 0 {
```
vkhoroz: Hmm... Our linter should have suggested to use `slices.Contains`.... Even if not - that's a good idea.

server/ui/web/templates/devices_list.html
```diff
@@ -5,10 +5,34 @@ <h2>{{.Title}}</h2>
...
-	    <th>Name</th>
-            <th>Created at</th>
-            <th>Last seen</th>
+            <th class="sortable">
+              {{ if eq .Sort "uuid-asc" }}
+                <a href="/devices?sort=uuid-desc" title="Sort descending">UUID <span>↓</span></a>
+              {{ else }}
+                <a href="/devices?sort=uuid-asc" title="Sort ascending">UUID <span>↑</span></a>
+              {{ end }}
+            </th>
```
vkhoroz, line 16: I guess there are 3 states in each (sortable) column: sorted ascending, sorted descending, not used for sorting.

```suggestion
            <th class="sortable">
              {{ if eq .Sort "uuid-asc" }}
                <a href="/devices?sort=uuid-desc" title="Sort descending">UUID <span>↓</span></a>
              {{ else if .Sort "uuid-desc" }}
                <a href="/devices?sort=uuid-asc" title="Sort ascending">UUID <span>↑</span></a>
              {{ else }}
                <a href="/devices?sort=uuid-asc" title="Sort ascending">UUID</a>
              {{ end }}
            </th>
```

server/ui/web/templates/devices_list.html
```diff
@@ -5,10 +5,34 @@ <h2>{{.Title}}</h2>
...
+            <th class="sortable">
+              {{ if eq .Sort "name-asc" }}
+                <a href="/devices?sort=name-desc" title="Sort descending">Name <span>↓</span></a>
+              {{ else }}
+                <a href="/devices?sort=name-asc" title="Sort ascending">Name <span>↑</span></a>
```
vkhoroz: What about paging?
Do you think that re-sorting should reset the paging to zero (like now)?

server/ui/web/templates/devices_list.html
```diff
@@ -72,7 +72,7 @@ <h2>{{.Title}}</h2>
             <a href="/devices?page={{sub .Page 1}}{{if .Sort}}&amp;sort={{.Sort}}{{end}}" role="button">&larr; Previous</a>
           {{ end }}
         </div>
-        <span>Page {{add .Page 1}}</span>
+        <span>Page {{add .Page 1}} of <a href="/devices?page={{sub .TotalPages 1}}{{if .Sort}}&amp;sort={{.Sort}}{{end}}">{{.TotalPages}}</a></span>
```
vkhoroz: It'd be nice to also have a way to click back to the first page when offset is non-zero.

vkhoroz (commented):
Looks nice.

server/ui/web/templates/devices_list.html
doanac: I think that's the element of least surprise

cli/subcommands/devices/list.go
doanac: fixed in force push to 100

cli/subcommands/devices/list.go
doanac: I updated this a bit. In the force push it now looks like:
image

Where previous/first and next/last are optionally shown

storage/api/api_storage_test.go
doanac: fixed in force push

cli/subcommands/devices/list.go
doanac: Oops - that reply was meant for another comment, but now you see.

I've also updated the CLI to show:
```
A total of 20 pages of devices available. Use '--page 1' for the next page.
```

cli/subcommands/devices/list.go
doanac: fixed in force push

server/ui/web/templates/devices_list.html
doanac: This is a bit of a UX problem. I poked around and made a small change. It now looks like:
image

The "up-down" arrow indicates a column can be sorted. Once sorted, it shows the direction its sorted it and the tooltip when you hover over says by clicking it will reverse that.

server/ui/web/templates/devices_list.html
doanac: As noted in the wrong reply, this logic was updated

server/ui/web/templates/devices_list.html
vkhoroz: That's brilliant. Thanks!

cli/subcommands/devices/list.go
vkhoroz: Can we make 50 to be on par with the UI?
I'm sorry to sound annoying. I typed the above comment before reading your UI code.

server/ui/web/handlers_devices.go
```diff
@@ -14,23 +17,52 @@ import (
 )
 
 func (h handlers) devicesList(c echo.Context) error {
+	page, _ := strconv.Atoi(c.QueryParam("page"))
+	if page < 0 {
+		page = 0
+	}
```
vkhoroz, line 23: Just commenting here to remind you to check how the UI is affected by the numbering from zero vs one, according to the comment I posted for the CLI.

cli/subcommands/devices/list.go
doanac: good catch. fixed in force push

cli/subcommands/devices/list.go
```diff
@@ -46,6 +49,7 @@ func init() {
 	DevicesCmd.AddCommand(listCmd)
 	listCmd.Flags().StringP("columns", "", "uuid,target,last-seen",
 		"Comma-separated list of columns to display (available: "+colmnsStr+")")
+	listCmd.Flags().IntP("page", "p", 0, "Page number to display (0-indexed)")
```
vkhoroz: I know that for us programmers numbering from zero is the most natural, and that French parliament declared that natural numbers start from zero.
But, for a regular person, numbering from one is preferred, especially given the majority of sites/programs doing that way.
When someone sees `Use '--page 1' for the next page` when that next page is the second one - they might be confused.

It's a bit of annoyance for coding. Still, it is worth to have a less confusing UX.
I missed this in the first review iteration, sorry.

vkhoroz (approved):
This is an LGTM now.

I only have one comment I missed initially.

storage/api/api_storage.go
vkhoroz: I guess you object to this comment (that's fine - you have your reasons).
Only responding to check if you didn't forget about it.

cli/subcommands/devices/list.go
doanac: this is indeed confusing and produces weirdness. Look at what the URL says versus what the page text shows:
image

storage/api/api_storage.go
doanac: I'd rather wait until we actually want to do such a thing.

cli/subcommands/devices/list.go
vkhoroz: My counter-argument for confusion is that the "first" link in the header currently points to page zero (not one).
As confusing is `Showing page 0 of 40 pages` or `Showing page 39 of 40 pages`. It requires a mental pain to get that page 39 is the last one.

That only means that we need to number since one on the API side too.
Obviously `page 2` in a footer should correspond to `page=2` in the URL; but it also needs to be the 2nd page, not 3rd.
FWIW our FoundriesFactory APIs are numbering since one for paging.

doanac: I've pushed 8ffc8f8cafeaa3e69b6fc1be902940e125a56611 to make a 1-based index

cli/subcommands/devices/list.go
```diff
@@ -61,7 +61,7 @@ func init() {
 	DevicesCmd.AddCommand(listCmd)
 	listCmd.Flags().StringP("columns", "", "uuid,target,last-seen",
 		"Comma-separated list of columns to display (available: "+colmnsStr+")")
-	listCmd.Flags().IntP("page", "p", 0, "Page number to display (0-indexed)")
+	listCmd.Flags().IntP("page", "p", 1, "Page number to display (1-indexed)")
```
vkhoroz: That `(1-indexed)` thing sounds weird now. I'd rather remove that text, as we deferred to the naive numbering.

vkhoroz (approved):
LGTM

cli/subcommands/devices/list.go
doanac: yep - and the help text already indicates this:
```
  -p, --page int         Page number to display (default 1)
```
fixed in force-push

PR #137, opened 2026-04-15
Security improvements

doanac:
Running some Claude analysis on the project yielded some interesting findings. Nothing major but will help keep the noise down for future code scans.


auth/csrf.go
```diff
@@ -16,8 +16,7 @@ const CsrfCookieName = "dg-satellite-csrf"
 const CsrfHeaderName = "X-CSRF-Token"
 
 // SetCsrfCookie sets a CSRF cookie on the response. It should be called
-// when a new session is created. The cookie is NOT HttpOnly so that
-// JavaScript can read the value to include in request headers.
+// when a new session is created.
```
vkhoroz: This fits into one line now.

cli/config/config.go
```diff
@@ -82,11 +82,6 @@ func SaveConfig(configPath string, cfg *Config) error {
 		}
 	}
 
-	configDir := filepath.Dir(configPath)
-	if err := os.MkdirAll(configDir, 0755); err != nil {
-		return fmt.Errorf("failed to create config directory: %w", err)
-	}
-
```
vkhoroz, line 89: I have a related comment on this.

A problem I faced recently was that when I run `satctl login` for the first time - it complains that "the config file does not exist".
This is a real usability problem, as I had to look into the code for the default config location, and then create that file manually.
Not only that, I had to put smth like `{}` into that file content.

IMHO we need to create a config file if it does not yet exist during a `satctl login` command, at least at our default location.
For non-default location, we might ask a user if they want to create it.

Now, IIUC creating a file sometimes requires creating a directory.
So we still need create it too, and set proper permissions, but only when the user authorizes (not for all commands implicitly).

vkhoroz (approved):

cli/config/config.go
doanac: weird. it looks like it should handle this: https://github.com/foundriesio/dg-satellite/blob/4bfdf0903cda08f42486a806c81a5aac8b098ad1/cli/subcommands/login/cmd.go#L48-L57

I'm hoping to improve some UX of this in the future by adding oauth2 device-flow support. I have a branch that does this to some extent, but we have more pressing things at this moment.

auth/csrf.go
doanac: fixed in force-push

cli/config/config.go
doanac: I found the bug - https://github.com/foundriesio/dg-satellite/blob/main/cli/config/config.go#L39

doanac: 30bb7aefc0ce20c59f4d7499ed798d9f4821ef0c

vkhoroz: Thanks. I think that's it.

I was getting that error, and after I created an empty file - I was getting unmarshal error.
Now I think your login code should handle the no-exist error correctly, and a user should have no incentive to create an empty file.

PR #138, opened 2026-04-16
storage: Fix possible double yield in iterator

doanac:
The yield on line 206 could return false but we'd still wind up calling the yield on line 200.

### DIFF
--- a/storage/file.go
+++ b/storage/file.go
@@ -204,6 +204,7 @@ func (s baseFsHandle) readFileLines(name string, ignoreNotExist bool, infinitySt
 			}
 			if err = scanner.Err(); err != nil {
 				yield("", err)
+				return
 			}
 			if infinityStop != nil {
 				// Tail functionality - simply re-create the scanner with the same fd after some time.


doanac:
VSCode has complained to me about this for ages. I think this is what you intended for this loop.

storage/file.go
```diff
@@ -203,7 +203,9 @@ func (s baseFsHandle) readFileLines(name string, ignoreNotExist bool, infinitySt
 				}
 			}
 			if err = scanner.Err(); err != nil {
-				yield("", err)
+				if !yield("", err) {
+					return
+				}
```
vkhoroz: I think we should return unconditionally. There is no way to continue after an error, as we reuse the same file descriptor.
IIUC the caller is not willing to continue here as well; but it's better to have a safeguard in both places, just in case if the caller was dumb enough to not handle an error.

vkhoroz (approved):
Thanks for this fix.

storage/file.go
doanac: updated in force

PR #140, opened 2026-04-20
fixup: quick fixes for watchdog, tests, and linter

vkhoroz:


doanac (approved):

go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/mattn/go-sqlite3 v1.14.32
 	github.com/spf13/cobra v1.10.2
 	github.com/stretchr/testify v1.11.1
+	github.com/swaggo/swag v1.16.6
```
doanac, line 15: This actually causes problems for me. We don't use swag in our actual code. Its only used to generate docs for releases and PRs

doanac: So now my editor is warning me about needing to run "go mod tidy"

vkhoroz: Without this, when I run `golangci-lint run`, I get a plenty of errors referencing to the "not imported" swaggo.
But I guess, that might be because I run `make swagger` at some point in time.

Need to think how to resolve this.

vkhoroz: Actually, having swagger in `go.mod` is good, as otherwise its version is "free to be anything", allowing for the supply chain attacks.

We currently fetch latest swaggo/swag during the release process.
I guess that we can do the anonymous import of it somewhere, use `go mod download` inside release, and let the go mod machinery do the proper job.

WDYT?

vkhoroz: I think I'll leave this up to you, as tbh I've no quick idea how to resolve this.

I was only trying to get linter pass locally, and as you can see github actions passed too.
So, we're both at the "it works at my laptop" kind of a thing now - but either way causes a problem to one of our laptops.

I think I can try to delete all auto-generated swagger files on my laptop, and try to never run `make swagger` again.
But that makes it inconvenient to check if swagger generates well locally.

vkhoroz: @doanac this is the idea: https://github.com/foundriesio/dg-satellite/pull/142

doanac: I think the reason you see swaggo stuff is because you ran the Makefile w/o this update: 58f1813f9d10b32c8c2bda674b9ad5eaaa5e6f12

With that change, we no longer produce a file that references swaggo

vkhoroz: Yes, you're right: I run it a while ago, before that change.

Now I removed `doc/swagger`, and after re-run, there is no more dependency onto swaggo.

https://github.com/foundriesio/dg-satellite/pull/143

PR #141, opened 2026-04-20
ui: Handle TUF issues gracefully in updates view

doanac:
I see this in my instance where I'm doing custom CI for fioup and only need a targets.json file to exist. We fail to parse *all* the TUF metadata and don't even show the content about the update.

### DIFF
--- a/storage/file_updates.go
+++ b/storage/file_updates.go
@@ -5,6 +5,7 @@ package storage
 
 import (
 	"crypto/rand"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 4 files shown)


storage/file_updates.go
```diff
@@ -40,6 +41,42 @@ func (s *updatesFsHandleWrap) init(root string) {
...
+	if err = json.Unmarshal(content, &targets); err != nil {
+		return fmt.Errorf("error parsing targets.json: %w", err)
+	}
+
+	for _, t := range targets.Signed.Targets {
+		if slices.Contains(t.Custom.Tags, tag) {
+			return nil
+		}
+	}
+	fmt.Println("ANDY invalid", string(content))
```
vkhoroz: debug detected

storage/file_updates.go
```diff
@@ -40,6 +41,42 @@ func (s *updatesFsHandleWrap) init(root string) {
...
+func checkUpdateTargets(targetsPath, tag string) error {
+	content, err := os.ReadFile(targetsPath)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("missing required targets.json file in tuf directory")
+		}
+		return fmt.Errorf("error reading targets.json: %w", err)
+	}
+
+	var targets tufTargetsSantity
```
vkhoroz: If you only need this struct for validation in a single place, you might declare the entire struct right here as an anonymous type:
```
var targets struct {
}
```

storage/file_updates.go
```diff
@@ -73,6 +110,11 @@ func (s updatesFsHandleWrap) SaveUpload(tag, update string, payload io.Reader, o
 					return fmt.Errorf("%w: must contain %q and/or %q directory",
 						ErrInvalidUpdate, UpdatesOstreeDir, UpdatesAppsDir)
 				}
+
+				path := filepath.Join(root, txDir, "unpacked/tuf/targets.json")
+				if err := checkUpdateTargets(path, tag); err != nil {
+					return fmt.Errorf("%w: %v", ErrInvalidUpdate, err)
+				}
```
vkhoroz, line 115: I'm not a great fan of deep diving into the validation of the updates tarball, as there are 100500 ways how it may be screwed up.

But, if we agree to only stick of a handful of trivial things like this one - let it be as you wish.

doanac: yep. This one is just so easy to make. We give the user the impression they can give the update any tag name they want. I fell victim for it and it took me 30 minutes to understand why.

storage/file_updates.go
doanac: fixed in force push

storage/file_updates.go
doanac: fixed in force push

vkhoroz (approved):

PR #142, opened 2026-04-20
swagger: let go mod control the swaggo/swag version properly

vkhoroz:
Currently, swagger is only imported after `make swagger` command was run. Hence, depending onto the user's environment, go mod and linter may require or reject it inside go.mod. Anonymous import lets the go.mod do the dependency work properly for us.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -80,7 +80,7 @@ jobs:
         with:
           go-version: "1.24"
       - name: Install swag
-        run: go install github.com/swaggo/swag/cmd/swag@latest
+        run: go install github.com/swaggo/swag/cmd/swag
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -80,7 +80,7 @@ jobs:
         with:
           go-version: "1.24"
       - name: Install swag
-        run: go install github.com/swaggo/swag/cmd/swag@latest
+        run: go install github.com/swaggo/swag/cmd/swag
```
vkhoroz, line 83: I didn't test this line.

Documentation says that it should get the version to install from the go mod.

vkhoroz:
See https://github.com/foundriesio/dg-satellite/pull/140#discussion_r3113517797

PR #144, opened 2026-04-22
Introduce rate limits to authentication related APIs

doanac:
This is an extension of PR #119 to achieve the same goals but in a better way.

### DIFF
--- a/auth/ratelimiter.go
+++ b/auth/ratelimiter.go
@@ -19,6 +19,10 @@ import (
 var errTooManyRequests = errors.New("rate-limit exceeded")
 var errTooManyBadAuthOps = errors.New("too many bad authentication operations")
 
+// badAuthRate is the token-bucket refill rate for bad-auth tracking,
+// set to 1 token per 60 seconds (i.e. 1 per minute).
+const badAuthRate = rate.Limit(1.0 / 60)
+
 func NewRateLimiter(cfg storage.RateLimitConfig) *authRateLimiter {
 	if cfg.AttemptsPerSecond <= 0 {
 		cfg.AttemptsPerSecond = 2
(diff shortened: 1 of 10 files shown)


doanac:
@vkhoroz - this is closer to what we've been talking about. I have a feeling it still needs a little tweaking, but its finally looking usable.

auth/ratelimiter.go
```diff
@@ -0,0 +1,163 @@
...
+	// Check the per-second rate limit; if exceeded, block the IP
+	rv, rvExists := rl.rateLimits[identifier]
+	if !rvExists {
+		rv = &rateLimiter{
+			Limiter: rate.NewLimiter(rate.Limit(rl.attemptsPerSecond), rl.attemptsPerSecond),
+		}
+		rl.rateLimits[identifier] = rv
+	}
+	rv.lastSeen = now
+	if !rv.Allow() {
```
vkhoroz: Why do we need two `rateLimit` classes and two maps?
I guess one would suffice and make this both simpler and faster:
```suggestion
// Book keeping of per-IP rate limiters.
v, exists := rl.limiters[identifier]
if !exists {
	const MinutesPerSecond float64 = 1.0 / 60.0
	v = &limiter{
			AuthLimiter: rate.NewLimiter(rate.Limit(MinutesPerSecond), rl.badAuthLimit),
			RateLimiter: rate.NewLimiter(rate.Limit(rl.attemptsPerSecond), rl.attemptsPerSecond),
		}
	}
	rl.limiters[identifier] = v
}
v.lastSeen = now

// Check if this IP is already blocked due to bad auth operations
if now.Before(v.blockUntil) || v.Tokens() <= 0 {
	return v.blockReason
}
// Check if auth attempt rate is too high
if !v.RateLimiter.Allow() {
	// No need to have that auth limiter lookup logic below again - for the third time already.
	// Just set the block reason and return an error.
```

auth/ratelimiter.go
```diff
@@ -0,0 +1,163 @@
...
+		if !exists {
+			r := float64(1) / 60
```
vkhoroz: This deserves a constant.

auth/ratelimiter.go
```diff
@@ -19,6 +19,53 @@ import (
...
+// put inserts or updates a key in front.
+func (g *generationMap[V]) put(key string, v V) {
+	g.front[key] = v
+}
+
+// sweep rotates generations if sweepAge has elapsed since the last swap.
+func (g *generationMap[V]) sweep(now time.Time) {
+	if now.Sub(g.lastSwap) >= g.sweepAge {
+		g.back = g.front
+		g.front = make(map[string]V, len(g.back)/2)
```
vkhoroz: Can we use `(len(g.back) + len(g.front)) / 2` here?

auth/ratelimiter.go
```diff
@@ -19,6 +19,53 @@ import (
 var errTooManyRequests = errors.New("rate-limit exceeded")
 var errTooManyBadAuthOps = errors.New("too many bad authentication operations")
 
+// generationMap is a two-generation map that provides O(1) garbage collection.
+// Active entries are promoted to `front` on access. Periodically, `front` is
+// rotated to `back` and a fresh `front` is allocated, discarding the old `back`.
```
vkhoroz: Thanks for this.
I guess we can extract these low-level classes into e.g. `auth/generations.go`

vkhoroz (approved):
This looks quite nice.

auth/ratelimiter.go
doanac: Yep - code is much cleaner. I've just appened a commit to the end rather than set it at first and then rework a bunch of merge conflict.

auth/ratelimiter.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func (rl *authRateLimiter) FlagBadOperation(c echo.Context) {
+	rl.mutex.Lock()
+	defer rl.mutex.Unlock()
+
+	identifier := c.RealIP()
+	v := rl.getOrCreate(identifier)
+	v.lastSeen = time.Now()
+	if !v.authLimit.AllowN(v.lastSeen, 1) {
```
vkhoroz: I might be wrong.... yet it looks like the `v.lastSeen` is never effectively used anymore.
The only place were I still see it in use "nominally" is here. But we could've used a local variable instead.

auth/ratelimiter.go
```diff
@@ -0,0 +1,138 @@
...
+		v.blockUntil = time.Now().Add(rl.badAuthBlockDuration)
```
vkhoroz: I think we can reuse the "now" from line 106 here too.

vkhoroz (approved):
LGTM

auth/ratelimiter.go
doanac: nice catch. fixed in force push

PR #146, opened 2026-04-23
Device search

doanac:
Add the ability to filter/search for devices by their labels when listing.

It builds on the idea of filters that look like `<label>.<compartor>.<value>`. So to filter for devices who's name contains "foo", you say `name.contains.foo`. These can be ANDed together for a query like: `name.contains.19 name.contains.2` which could return devices with names like `192`, `219`, `1932`, etc.

The auto-completion logic in the UI makes it a little easier to use for a first timer.

### DIFF
--- a/server/ui/api/handlers_devices.go
+++ b/server/ui/api/handlers_devices.go
@@ -37,6 +37,7 @@ type LabelsPutReq map[string]*string
 // @Description Requires scope: devices:read or devices:read-update
 // @Tags    Devices
 // @Param _ query DeviceListOpts false "Sorting options"
+// @Param label query []string false "Label filters in the format key.comparison.value (e.g. env.eq.production). Comparison operators: eq, ne, contains, ncontains. Multiple filters are ANDed together."
 // @Accept  json
 // @Produce json
 // @Success 200 {array} DeviceListItem
@@ -52,6 +53,12 @@ func (h *handlers) deviceList(c echo.Context) error {
 		return EchoError(c, err, http.StatusBadRequest, "Failed to parse list options")
 	}
 
+	if filters, err := parseLabelFilters(c.QueryParams()["label"]); err != nil {
+		return EchoError(c, err, http.StatusBadRequest, err.Error())
+	} else {
+		opts.LabelFilters = filters
+	}
+
 	devices, total, err := h.storage.DevicesList(opts)
 	if err != nil {
 		return EchoError(c, err, http.StatusInternalServerError, "Unexpected error listing devices")
@@ -59,6 +66,32 @@ func (h *handlers) deviceList(c echo.Context) error {
 
 	setPaginationHeaders(c, opts, total)
 	return c.JSON(http.StatusOK, devices)
+}
+
+// parseLabelFilters parses query parameters of the form "key.comparison.value"
+// into LabelFilter structs. The value may contain dots.
+func parseLabelFilters(params []string) ([]storage.LabelFilter, error) {
+	if len(params) == 0 {
+		return nil, nil
+	}
+	filters := make([]storage.LabelFilter, 0, len(params))
+	for _, p := range params {
+		// Split into at most 3 parts: key, comparison, value (value may contain dots)
+		parts := strings.SplitN(p, ".", 3)
+		if len(parts) != 3 {
+			return nil, fmt.Errorf("invalid label filter %q: expected format key.comparison.value", p)
+		}
+		f := storage.LabelFilter{
+			Label:      parts[0],
+			Comparison: storage.LabelComparison(parts[1]),
+			Value:      parts[2],
+		}
+		if err := f.Validate(); err != nil {
+			return nil, err
+		}
+		filters = append(filters, f)
+	}
+	return filters, nil
 }
 
 func setPaginationHeaders(c echo.Context, opts storage.DeviceListOpts, total int) {
--- a/server/ui/web/templates/devices_list.html
+++ b/server/ui/web/templates/devices_list.html
@@ -2,43 +2,81 @@
     <section class="content-section">
       <h2>{{.Title}}</h2>
 
+      <input type="search" id="labels" name="labels" placeholder="Search eg. name.contains.foo"> 
+      <p class="filter-help"><small>Comparisons: <code>eq</code> (equal), <code>ne</code> (not equal), <code>contains</code>, <code>ncontains</code> (not contains)</small></p>
+      <script>
+      // Populate the labels input with current label query params on page load
+      document.addEventListener('DOMContentLoaded', function() {
+        var labelsInput = document.getElementById('labels');
+          // Set input value from query params
+          var params = new URLSearchParams(window.location.search);
+          var labelsArr = [];
+          params.forEach(function(value, key) {
+            if (key == "label") {
+              labelsArr.push(value);
+            }
+          });
+          if (labelsArr.length > 0) {
+            labelsInput.value = labelsArr.join(' ');
+          }
+
+          // When Enter is pressed in the labels input, update query params and reload
+          labelsInput.addEventListener('keydown', function(e) {
+            if (e.key === 'Enter') {
+              e.preventDefault();
+              var labels = labelsInput.value.trim();
+              var params = new URLSearchParams(window.location.search);
+              [...params.keys()].forEach(k => { if (k == "label") params.delete(k); });
+              // Add new label params from space-delimited input
+              if (labels.length > 0) {
+                labels.split(/\s+/).forEach(function(label) {
+                  params.set("label", label);
+                });
+              }
+              // Keep other params (sort, page, etc)
+              window.location.search = params.toString();
+            }
+          });
+      });
+      </script>
+
       <table class="striped">
         <thead>
           <tr>
             <th class="sortable">
               {{ if eq .Sort "uuid-asc" }}
-                <a href="/devices?sort=uuid-desc" title="Sorted ascending, click for descending">UUID <span class="sort-active">▲</span></a>
+                <a href="/devices?sort=uuid-desc{{labelParams .LabelFilters}}" title="Sorted ascending, click for descending">UUID <span class="sort-active">▲</span></a>
               {{ else if eq .Sort "uuid-desc" }}
-                <a href="/devices?sort=uuid-asc" title="Sorted descending, click for ascending">UUID <span class="sort-active">▼</span></a>
+                <a href="/devices?sort=uuid-asc{{labelParams .LabelFilters}}" title="Sorted descending, click for ascending">UUID <span class="sort-active">▼</span></a>
               {{ else }}
-                <a href="/devices?sort=uuid-asc" title="Sort by UUID">UUID <span class="sort-idle">⇅</span></a>
+                <a href="/devices?sort=uuid-asc{{labelParams .LabelFilters}}" title="Sort by UUID">UUID <span class="sort-idle">⇅</span></a>
               {{ end }}
             </th>
             <th class="sortable">
               {{ if eq .Sort "name-asc" }}
-                <a href="/devices?sort=name-desc" title="Sorted ascending, click for descending">Name <span class="sort-active">▲</span></a>
+                <a href="/devices?sort=name-desc{{labelParams .LabelFilters}}" title="Sorted ascending, click for descending">Name <span class="sort-active">▲</span></a>
               {{ else if eq .Sort "name-desc" }}
-                <a href="/devices?sort=name-asc" title="Sorted descending, click for ascending">Name <span class="sort-active">▼</span></a>
+                <a href="/devices?sort=name-asc{{labelParams .LabelFilters}}" title="Sorted descending, click for ascending">Name <span class="sort-active">▼</span></a>
               {{ else }}
-                <a href="/devices?sort=name-asc" title="Sort by Name">Name <span class="sort-idle">⇅</span></a>
+                <a href="/devices?sort=name-asc{{labelParams .LabelFilters}}" title="Sort by Name">Name <span class="sort-idle">⇅</span></a>
               {{ end }}
             </th>
             <th class="sortable">
               {{ if eq .Sort "created-at-asc" }}
-                <a href="/devices?sort=created-at-desc" title="Sorted ascending, click for descending">Created at <span class="sort-active">▲</span></a>
+                <a href="/devices?sort=created-at-desc{{labelParams .LabelFilters}}" title="Sorted ascending, click for descending">Created at <span class="sort-active">▲</span></a>
               {{ else if eq .Sort "created-at-desc" }}
-                <a href="/devices?sort=created-at-asc" title="Sorted descending, click for ascending">Created at <span class="sort-active">▼</span></a>
+                <a href="/devices?sort=created-at-asc{{labelParams .LabelFilters}}" title="Sorted descending, click for ascending">Created at <span class="sort-active">▼</span></a>
               {{ else }}
-                <a href="/devices?sort=created-at-asc" title="Sort by Created at">Created at <span class="sort-idle">⇅</span></a>
+                <a href="/devices?sort=created-at-asc{{labelParams .LabelFilters}}" title="Sort by Created at">Created at <span class="sort-idle">⇅</span></a>
               {{ end }}
             </th>
             <th class="sortable">
               {{ if eq .Sort "last-seen-asc" }}
-                <a href="/devices?sort=last-seen-desc" title="Sorted ascending, click for descending">Last seen <span class="sort-active">▲</span></a>
+                <a href="/devices?sort=last-seen-desc{{labelParams .LabelFilters}}" title="Sorted ascending, click for descending">Last seen <span class="sort-active">▲</span></a>
               {{ else if eq .Sort "last-seen-desc" }}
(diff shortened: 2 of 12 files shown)


doanac:
This addresses issue #146

doanac:
@vkhoroz - no big rush here. I let Claude take a stab at this today and I was pretty pleased with its results.

doanac:
Screenshots:
## no filter applied
image

## two filters applied
image

i can't catch a screenshot of autocomplete

storage/api/api_storage.go
```diff
@@ -213,24 +238,96 @@ func (s Storage) DevicesList(opts DeviceListOpts) ([]DeviceListItem, int, error)
...
+		}
+		// Use the JSON extract operator with a literal key path.
+		// The key is embedded as a JSON path literal; the value is parameterized.
+		jsonPath := "$." + f.Label
+		switch f.Comparison {
+		case LabelCmpEqual:
+			sb.WriteString(fmt.Sprintf(" AND labels ->> '%s' = ?", jsonPath))
+			args = append(args, f.Value)
+		case LabelCmpNotEqual:
+			sb.WriteString(fmt.Sprintf(" AND (labels ->> '%s' IS NULL OR labels ->> '%s' != ?)", jsonPath, jsonPath))
```
vkhoroz: I will take a deeper look at this later. Seems quite interesting.

One thing I've noticed immediately is that our indexes for name and group labels use the following `COALESCE(labels ->> '$.name', "")`. With that in mind, this new search code will not utilize those indexes as it is not the same. In order to utilize them, a field operation must take _exactly the same form_ verbatim.

Can you consider that in this code and try using e.g. sqlite's explain statement to check if indexes are used?

doanac: I was able to improve this by adding special logic for those two fields. Updated in force push

storage/api/api_storage.go
```diff
@@ -477,64 +585,6 @@ func (s *stmtDeviceGet) run(
...
 
-type stmtDeviceList storage.DbStmt
-
-func (s *stmtDeviceList) Init(db storage.DbHandle, orderBy string) (err error) {
-	s.Stmt, err = db.Prepare("apiDeviceList", fmt.Sprintf(`
-		SELECT
-			uuid, created_at, last_seen, target_name, tag, is_prod, json(labels)
-		FROM devices
-		WHERE deleted=false
-		ORDER BY %s LIMIT ? OFFSET ?`, orderBy),
```
vkhoroz, line 589: I think we should still use prepared statements for the basic (non-filtered) use case, which will be used in 80-95% of cases.
Fortunately, an implementation with prepared statements could be easily extended.

For example, changing the last two lines of the above statement gives you exactly what you need:
```
		WHERE deleted=false %s
		ORDER BY %s LIMIT ? OFFSET ?`, filterBy, orderBy),
```

Then, basic statements (where `filterBy` is empty) will be initialized just as before, while filtered queries will call Init+run ad-hoc.

Again, this already improves the non-filtered (most used) queries. But not only that....
That slightly more structured implementation allows to easily add a caching layer for prepared filtered statements in the future. Alas, there is a finite number of "filtering" combinations, so small that I bet their lazy initialization can easily fit into as small map as hundreds (or even dozens).
(If you wonder about a caching key - it could be e.g. a `sha1(filterBy+orderBy)`.

storage/api/api_storage.go
```diff
@@ -215,24 +240,107 @@ func (s Storage) DevicesList(opts DeviceListOpts) ([]DeviceListItem, int, error)
...
 		return nil, 0, err
 	}
 
 	return devices, total, nil
 }
 
+// buildLabelFilterSQL produces SQL WHERE clause fragments and args for label filters.
+// Each filter uses the JSONB extract operator (labels ->> '$.<key>') with parameterized values.
+// For "name" and "group" labels, it uses the indexed virtual columns directly.
+func buildLabelFilterSQL(filters []LabelFilter) (string, []any, error) {
```
vkhoroz, line 306: I think we need to move `LabelFilter` and `buildLabelFilterSQL` into a separate module like e.g. `db_filter.go`.
They are generic enough that eventually we'd like to extend them to non-label and/or non-device queries.

storage/api/api_storage.go
```diff
@@ -41,6 +42,41 @@ const (
...
 )
 
+// LabelComparison defines the comparison operator for label filters.
+type LabelComparison string
+
+const (
+	LabelCmpEqual       LabelComparison = "eq"
+	LabelCmpNotEqual    LabelComparison = "ne"
+	LabelCmpContains    LabelComparison = "contains"
+	LabelCmpNotContains LabelComparison = "ncontains"
```
vkhoroz, line 56: Just IMHO, from a lexical perspective we can use any character sequences for these comparisons.
From a usability perspective, operands like `name = abcd, group != foo, group ~= f*, name ~!= a?cd` are more intuitive than, and require as much backend coding as `name.eq.abcd, group.ne.foo, group.contains.f*, name.ncontains.a?cd`.

We can actually use words at the API level (iiuc easier to URL-escape), but I'd much prefer to use more intuitive querying at the UX level.
Another option is to wrap entire filter into e.g. base32 encoding before calling API, in which case anything can be inside it.

fwiw your `contains` really seems to work as `like` i.e. allows for a true pattern matching, in line with the widespread `~=` use and unlike a limited `contains` meaning.

Also, I'm a bit skeptical about the `ncontains` thing. Is this really what we want to support (i.e. are there real use cases for this)? If it is just to satisfy a rule of 4 - I'd remove it, as that kind of search is often associated with performance implications.

storage/api/api_storage.go
```diff
@@ -41,6 +42,41 @@ const (
...
+
+// LabelFilter defines a filter on a device's JSONB labels column.
+type LabelFilter struct {
+	Label      string
+	Value      string
+	Comparison LabelComparison
+}
+
+// validLabelKey ensures label keys only contain safe characters for JSON path embedding.
+var validLabelKey = regexp.MustCompile(`^[a-zA-Z0-9_\-]+$`)
```
vkhoroz, line 67: I'd really wish to allow `*?` characters for the match operators, where we use LIKE internally, but still require the first character to be non-glob.

storage/api/api_storage.go
```diff
@@ -215,24 +240,107 @@ func (s Storage) DevicesList(opts DeviceListOpts) ([]DeviceListItem, int, error)
...
+		// Use indexed virtual columns for "name" and "group" labels;
+		// fall back to JSON extraction for all other labels.
+		var col string
+		switch f.Label {
+		case "name":
+			col = "name"
+		case "group":
+			col = "group_name"
+		default:
+			col = fmt.Sprintf("labels ->> '$.%s'", f.Label)
```
vkhoroz, line 327: One other thing I'd like to support is `uuid`. It is not a label, but I often used it when debugging customer problems.
I do believe a `tag` field is yet another good example for filtering.

server/ui/api/handlers_devices.go
```diff
@@ -37,6 +37,7 @@ type LabelsPutReq map[string]*string
 // @Description Requires scope: devices:read or devices:read-update
 // @Tags    Devices
 // @Param _ query DeviceListOpts false "Sorting options"
+// @Param label query []string false "Label filters in the format key.comparison.value (e.g. env.eq.production). Comparison operators: eq, ne, contains, ncontains. Multiple filters are ANDed together."
```
vkhoroz, line 40: So, IIUC this requires the UI to pass multiple filters as multiple `label=x.y.z&label=a.b.c` query parameters (form style) and emphasizes on label filtering.

Maybe, we should consider a more broad filtering like `filter=x.y.z,a.b.c` extending beyond labels, and using a comma (or semicolon) as a separator for compactness (whatever separator the UI is currently using)?

server/ui/api/handlers_devices.go
```diff
@@ -61,6 +68,32 @@ func (h *handlers) deviceList(c echo.Context) error {
...
+			return nil, fmt.Errorf("invalid label filter %q: expected format key.comparison.value", p)
+		}
+		f := storage.LabelFilter{
+			Label:      parts[0],
+			Comparison: storage.LabelComparison(parts[1]),
+			Value:      parts[2],
+		}
+		if err := f.Validate(); err != nil {
+			return nil, err
+		}
```
vkhoroz, line 91: I think it would be beneficial to use some lexer, e.g. the [participle](https://pkg.go.dev/github.com/alecthomas/participle) library which allows defining the LabelFilter as below.

TBH. I'm not sure how much benefit any lexer would give us. A simple split  (as you did) seems sufficient to me, and in case if we switch to `=|~=|!=` kind of operators, a simple split would work well too.

```
var filterLexer = lexer.MustSimple([]lexer.SimpleRule{
	{Name: "Label", Pattern: `^[\w_]+`},
	{Name: "Operator", Pattern: `.eq.|.ne.|.contains.|.ncontains.`}, 
	{Name: "Value", Pattern: `[\w\.\-_]+$`},
})

type LabelFilter struct {
	Label      string          `parser:"@Label"`
	Comparison LabelComparison `parser:"@Operator"`
	Value      string          `parser:"@Value"`
}

// global initialization somewhere
parser, err := participle.Build[LabelFilter](participle.Lexer(filterLexer))
if err != nil { panic(err) }

// usage
input := "name.contains.abcd"
filter, err := parser.ParseString("", input)
```

server/ui/web/templates/devices_list.html
```diff
@@ -2,43 +2,81 @@
...
+            labelsInput.value = labelsArr.join(' ');
+          }
+
+          // When Enter is pressed in the labels input, update query params and reload
+          labelsInput.addEventListener('keydown', function(e) {
+            if (e.key === 'Enter') {
+              e.preventDefault();
+              var labels = labelsInput.value.trim();
+              var params = new URLSearchParams(window.location.search);
+              [...params.keys()].forEach(k => { if (k == "label") params.delete(k); });
```
vkhoroz: So, IIUC this deletes filter params from the existing window URL, right?
This deserves a comment I think.

server/ui/web/templates/devices_list.html
```diff
@@ -2,43 +2,81 @@
...
+              // Add new label params from space-delimited input
+              if (labels.length > 0) {
+                labels.split(/\s+/).forEach(function(label) {
+                  params.set("label", label);
+                });
+              }
+              // Keep other params (sort, page, etc)
+              window.location.search = params.toString();
```
vkhoroz: Hmm.... this is going to spam our API on each keydown.

IIRC there is a simple way to place a small delay to the event handler, so that the actual API call is only made after there was no subsequent keydown event for the next e.g. 300 or 500 milliseconds.
That's still responsive enough, but does not constantly refresh the page (both performance killer and UX flaw).

server/ui/web/templates/devices_list.html
```diff
@@ -2,43 +2,81 @@
...
+      });
+      </script>
+
       <table class="striped">
         <thead>
           <tr>
             <th class="sortable">
               {{ if eq .Sort "uuid-asc" }}
-                <a href="/devices?sort=uuid-desc" title="Sorted ascending, click for descending">UUID <span class="sort-active">▲</span></a>
+                <a href="/devices?sort=uuid-desc{{labelParams .LabelFilters}}" title="Sorted ascending, click for descending">UUID <span class="sort-active">▲</span></a>
```
vkhoroz, line 144: IIRC there is a way to hijack the page request and inject `{{labelParams .LabelFilters}}` on the fly, so that we don't need to remember copying it in so many places.

Smth along these lines:
```
document.addEventListener('click', function(e) {
  const link = e.target.closest('a');

  if (link && link.href && link.href.startsWith('/devices?')) {
    e.preventDefault();
    window.location.href = link.href+{{labelParams .LabelFilters}};
  }
});
```

It goes with the trade-off (truncated link in browser footer), but it saves quite some hurdle.

server/ui/web/templates/devices_list.html
```diff
@@ -2,42 +2,138 @@
...
-          var params = new URLSearchParams(window.location.search);
-          var labelsArr = [];
-          params.forEach(function(value, key) {
-            if (key == "label") {
-              labelsArr.push(value);
-            }
-          });
-          if (labelsArr.length > 0) {
-            labelsInput.value = labelsArr.join(' ');
+      (function() {
```
vkhoroz, line 10: This auto-completion is quite a thing, but I am scared by the maintenance cost.
Do you think this is what we really need at this stage?
Maybe, a simple help message above the filter input would do a better work?

vkhoroz (commented):
It looks quite interesting.

I'm almost approving this; but imho some things need shuffling around to decrease future maintenance costs.

storage/api/api_storage.go
doanac: you don't need to search for a UUID though?

server/ui/web/templates/devices_list.html
doanac: Claude can do the maintenance. This is too useful. I can't remember how to filter each time I use this. Depending how we do the query implementation, this might become even more needed.

storage/api/api_storage.go
doanac: its interesting. I just tried to get something on par with EC2. Let's get the simple thing going and see where that takes us.

storage/api/api_storage.go
doanac: we can change the filtering approach. I used EC2 GUI as something to look at. They include the same 4 operands and manage servers on the order of most people will have for devices. I'm testing with 10k devices and the time of the searches are imperceptible. I didn't see github or ec2 doing something like `name = a?cd`. They seem to lean on "contains" and "not contains" rather than taking on this complexity.

EC2's API is a little more like you refer to. eg `;tag:Name=!:og` is a search for VMs who's tag "Name" does not contain "og". Searching a not equal felt ugly `tag:Name=!%5C=og`. 

Pick a query language you want and I'll do it. The current one just seemed the easiest to write. That said, I'm just going to take your input and tell Claude to sort it out.

Some other options(`%3d` is urlencode of `=`) that are basically querying on "name equal" and "group not equal":

Messy to parse:  `filter=name%3Dfoo,group!%3Dbar`
Close to AWS: `tag:name=foo&tag:group=!%3Dbar`

vkhoroz: I think I made a wrong emphasize.

I don't quite care about how the API or URL is structured, but more about how the user input is.

I was proposing simple `=, !=, ~=` because that's how GCP console does search (e.g. for logging at al), and that's the most naive syntax any engineer understands, without a need to remember fancy keywords.

As for the URL itself, I'd simply make it opaque, e.g. using base32 _for the entire filter query_. I see no value in sending a filter verbatim in the URL. Often, APIs wrap the entire form data into a single `&state=abbracadabbraunreadablebase32pieceofjunk` kind of a thing, and tbh I like it so much more than all those fancy notations which make a life of a programmer harder by trying to be pseudo-"compliant".

storage/api/api_storage.go
vkhoroz: This makes sense.

I was just thinking: why got for EC2 way if FoundriesFactory way was already reacher? Why not GCP way? Or why not sqlite way?

That said, feel free to ignore. Contains is already powerful enough to solve 99% of use cases.

storage/api/api_storage.go
vkhoroz: Customer tickets usually contained a uuid or a set of uuids.
But, sometimes there were typos, and in cases like that I was searching by the uuid start.

storage/api/api_storage.go
doanac: I'm not against its idea and nomenclature, but GCP log search is quite a different beast than device search; at least if you compare to things like EC2 and GCP compute. GCP compute and github issues don't even really support search - its basically exact label matching. It makes me worry we are verging on overkill. Except, GCP compute and github issue search **infuriates** me.

I think the middle ground here would be to have the search be something entered into a text box (and also have a query helper like I've done) display something like `name = foo group != bar location ~= blah location !~` to cover equal, not equal, contains, and not contains. We can take that string and urlencode it to something like: `?filter=name%3Dfoo+group%21%3Dbar+location~%3Dblah`

vkhoroz: On our last comments we seem to be stating almost the same thing about a "potential middle ground syntax".
If you wish to url-encode instead of base32, I'm just fine.

Three things I would bother specifically:
- I don't see a use case for "not contains"; even "not equal" looks rather synthetic, but at least intuitive and marginally valuable.
- We do allow a dot inside label values (good, not need to change) - that's something to take in account when we talk about the syntax.
- Visual comfort of the user, i.e. seeing familiar things without clobbering.
   I bet nothing is more familiar than a school-level equality/inequality operators; and also school level (albeit less common) equivalence/approximation (~).  These are also widely used by JIRA/GCP/Databases, while GitHub/AWS took their unique paths.

vkhoroz: fwiw if we follow the AWS style, we might get into the licensing trouble. I'm not sure if their specific syntax is a common knowledge or patent-protected. Just something to keep in mind.

doanac: note: I'm not following AWS style - I found it confusing. I'm suggesting to support the same search operations as them. I've found them handy with my large fleet and surely they invested in small fortune in researching what would satisfy their users.

PR #147, opened 2026-04-27
web: Add help text when device has no assigned update

doanac:


doanac:
This helps with the usability. We were showing nothing for the assigned update. This change produces:
image

vkhoroz (approved):

PR #148, opened 2026-04-28
Device status updates

doanac:
This adds function we have in our current backend and product that was missing from the satellite server for showing a device's status when appropriate.

### DIFF
--- a/storage/api/api_storage.go
+++ b/storage/api/api_storage.go
@@ -97,6 +97,8 @@ type Device struct {
 	HwInfo  string `json:"hardware-info"`
 	NetInfo string `json:"network-info"`
 
+	Status *DeviceStatus `json:"status,omitempty"`
+
 	storage Storage
 }
 
@@ -268,8 +270,34 @@ func (s Storage) DeviceGet(uuid string) (*Device, error) {
 		return nil, err
 	}
 
+	// Find the most recent update event and derive device status
+	eventFiles, err := s.fs.Devices.ListFiles(d.Uuid, storage.EventsPrefix, true)
+	if err != nil {
+		return nil, err
+	}
+	if len(eventFiles) > 0 {
+		content, err := s.fs.Devices.ReadFile(d.Uuid, eventFiles[len(eventFiles)-1])
+		if err != nil {
+			return nil, err
+		}
+		content = strings.TrimSpace(content)
+		lastLine := content[strings.LastIndexByte(content, '\n')+1:]
+		var evt DeviceUpdateEvent
+		if err := json.Unmarshal([]byte(lastLine), &evt); err != nil {
+			return nil, fmt.Errorf("unexpected error unmarshalling event json: %w", err)
+		}
+
+		if !slices.Contains(clearingEventTypes, evt.EventType.Id) || evt.Event.Success == nil || !*evt.Event.Success {
+			// only share status if its interesting (ie not a successful update)
+			status := evt.ParseStatus()
+			d.Status = &status
+		}
+	}
+
 	return &d, nil
 }
+
+var clearingEventTypes = []string{"EcuInstallationCompleted", "CertRotationCompleted", "MetadataUpdateCompleted"}
 
 func (s Storage) ListUpdates(tag string, isProd bool) (map[string][]string, error) {
 	return s.getRolloutsFsHandle(isProd).ListUpdates(tag)
--- a/storage/models.go
+++ b/storage/models.go
@@ -3,7 +3,9 @@
 
 package storage
 
-import "regexp"
+import (
+	"regexp"
+)
 
 // DeviceUpdateEvent represents update events that devices send the
 // device-gateway.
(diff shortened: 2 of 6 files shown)


doanac:
The UI has tooltips that show target name and device time when you hover, but my screen capture tool can't get those
# UI when an problem is present
image

# UI when things are in progress
image

# UI when nothing is happening
image

# CLI example:
```
Status:       Installation completed; failed; 2025-11-06T18:21:26Z; intel-corei7-64-lmp-137
```

doanac:
@mike-sul @vkhoroz - I noticed this function was missing. Rather than trying to update the DB while events are spewing in, I thought it might be easier and more flexible for the future if we computed this value at runtime. That said, my logic could be invalid for a handful of reasons. Let me know what you think

storage/models.go
```diff
@@ -36,42 +36,59 @@ type DeviceStatus struct {
...
+		"EcuDownloadCompleted":     "Download completed",
+		"EcuInstallationStarted":   "Installation started",
+		"EcuInstallationApplied":   "Installation applied",
+		"EcuInstallationCompleted": "Installation completed",
+		"CertRotationStarted":      "Certificate rotation started",
+		"CertRotationCompleted":    "Certificate rotation completed",
+	}
+
 	var status string
+	update := e.Event.CorrelationId
```
vkhoroz: This variable is never used de facto.
(It is used inside one `case` statement, where a bool could be used instead).

storage/models.go
```diff
@@ -36,42 +36,59 @@ type DeviceStatus struct {
...
+		}
+		if e.Event.Success != nil {
+			if !*e.Event.Success {
+				status += "; failed"
+			} else if update == "" {
+				// If meta update, installation, or certificate rotation succeeds, then empty status which indicates success
+				clearingEvent = true
+			} else {
+				status += "; succeeded"
+			}
```
vkhoroz: This looks wrong.
For the `EcuInstallationCompleted` at al, we still need to append the `; succeeded` thing.

That `clearing` flag is kind of independent, and can be set separately:
```
// global
clearingEventTypes := []string{"EcuInstallationCompleted", "CertRotationCompleted", "MetadataUpdateCompleted"}
// inside this case statement
clearingEvent = slices.Contains(clearingEventNames, e.EventType.Id)
```

Nevertheless, just IMHO, this "clearing flag" is rather misleading at this level, as it is only needed for a "special" use case inside the device API, and confusing for the events API.
So, I'd rather examine the `e.EventType.Id` inside the device API directly.

storage/models.go
```diff
@@ -36,42 +36,59 @@ type DeviceStatus struct {
...
+	idToStatus := map[string]string{
+		"MetadataUpdateCompleted":  "Metadata update completed",
+		"EcuDownloadStarted":       "Download started",
+		"EcuDownloadCompleted":     "Download completed",
+		"EcuInstallationStarted":   "Installation started",
+		"EcuInstallationApplied":   "Installation applied",
+		"EcuInstallationCompleted": "Installation completed",
+		"CertRotationStarted":      "Certificate rotation started",
+		"CertRotationCompleted":    "Certificate rotation completed",
+	}
```
vkhoroz: Please, make it a global variable, so that it is not re-initialized (in a heap) on each function call.

storage/models.go
vkhoroz, line 83: In this new implementation `status` is never empty.

storage/api/api_storage.go
```diff
@@ -268,6 +270,34 @@ func (s Storage) DeviceGet(uuid string) (*Device, error) {
...
+		for i := len(lines) - 1; i >= 0; i-- {
+			if len(lines[i]) == 0 {
+				continue
+			}
+			var evt DeviceUpdateEvent
+			if err := json.Unmarshal([]byte(lines[i]), &evt); err != nil {
+				return nil, fmt.Errorf("unexpected error unmarshalling event json: %w", err)
+			}
+			if status := evt.ParseStatus(); status != nil {
+				if !status.ClearingEvent {
```
vkhoroz: This is the place: I think we could check `e.EventType.Id` here, and simplify the `ParseStatus` logic.

Actually, I see no harm if we always return the status, even if it is a "final/success" one.
I'm not even sure if we need to somehow hint the UI about the "final".

storage/models.go
```diff
@@ -36,42 +36,59 @@ type DeviceStatus struct {
...
+	case "EcuDownloadCompleted", "EcuInstallationCompleted", "CertRotationCompleted", "MetadataUpdateCompleted":
+		if e.EventType.Id != "EcuDownloadCompleted" {
+			update = ""
+		}
+		if e.Event.Success != nil {
+			if !*e.Event.Success {
+				status += "; failed"
+			} else if update == "" {
+				// If meta update, installation, or certificate rotation succeeds, then empty status which indicates success
+				clearingEvent = true
```
mike-sul: I guess we don't want to reset the `status` field here as we do in DG https://github.com/foundriesio/device-gateway/blob/b6560d88ae15cd555e1fabd05c62bd6caaaa4071/storage/types.go#L321 because we need the status value for the logic that infers/deduces status by parsing events?

storage/api/api_storage.go
```diff
@@ -268,6 +270,34 @@ func (s Storage) DeviceGet(uuid string) (*Device, error) {
...
+		lines := strings.Split(content, "\n")
+		for i := len(lines) - 1; i >= 0; i-- {
+			if len(lines[i]) == 0 {
+				continue
+			}
+			var evt DeviceUpdateEvent
+			if err := json.Unmarshal([]byte(lines[i]), &evt); err != nil {
+				return nil, fmt.Errorf("unexpected error unmarshalling event json: %w", err)
+			}
+			if status := evt.ParseStatus(); status != nil {
```
mike-sul: Why do we need to parse update events again to deduce/derive the device status if the events were parsed on arrival and the deduced status was written to a file system in https://github.com/foundriesio/dg-satellite/blob/5205696aa9af72e109d19103e8384b1b148d727e/storage/gateway/storage.go#L130. 
I guess there is something that does not allow to extract the status from `d.storage.fs.Updates.Ci.Logs` or `d.storage.fs.Updates.Prod.Logs`?

vkhoroz (commented):
hmm....
For some weird reason my comments did not get published yesterday.

storage/api/api_storage.go
doanac: It was a little easier to code and I figured if we could deduce something rather than write it out, it might make life easier.

I don't have a strong opinion.

storage/models.go
doanac: We need the status when we call this for adding to the rollouts/updates logs. We don't need it when we are trying to deduce the device status.

storage/models.go
doanac: it will be empty if an unknown event type is passed to the function

storage/models.go
doanac: sure - i was trying to keep it as close to the version in the device-gateway as possible to make the first pass at reviewing this easier.

storage/api/api_storage.go
mike-sul: It looks to me like this code writes the status anyway. If I read the code correctly, it deduces updates status out of update events it receives, then writes the status to a file system. After, that, the new code in this PR reads update events, deduces and status again, while it is already stored on file system.

storage/models.go
doanac: but when taken into account with your other comments, this will wind up being impossible.

storage/api/api_storage.go
doanac: we are writing one type of status to the updates rollout log. This "device status" is slightly different.

We can deduce this device status and store it (file or db) during event intake. I suspect we will be trading negligible negligible API side read performance with negligible device side write performance.

I'll add a commit to the end of this PR that does that and you and volodya can decide which you prefer

mike-sul: Ah, ok, I thought this is the same "status".

If it is so, I am fine with either options, just don't have enough understanding of the dg-sat source code to form an opinion about it.

storage/models.go
mike-sul: Do we have an issue with performance? If not, then keeping that map inside of the function makes code slightly more readable, IMHO.

storage/models.go
doanac: I reworked this to shuffle logic around. I think the new version is more inline with what you'll be wanting

storage/api/api_storage.go
```diff
@@ -268,9 +270,40 @@ func (s Storage) DeviceGet(uuid string) (*Device, error) {
...
+	if len(eventFiles) > 0 {
+		content, err := s.fs.Devices.ReadFile(d.Uuid, eventFiles[len(eventFiles)-1])
+		if err != nil {
+			return nil, err
+		}
+		lines := strings.Split(content, "\n")
+		for i := len(lines) - 1; i >= 0; i-- {
+			if len(lines[i]) == 0 {
+				continue
+			}
```
vkhoroz: I was thinking if there is any way to avoid a split and loop here, as we only really need the last non-empty line.

So, here is how that can be done:
```
content = strings.TrimSpace(content)
if len(content) > 0 {
	lastLine := content[strings.LastIndexByte(content, '\n')+1:]
	// lastLine is warranted to contain the last event, so there is no need for a loop.
    ... // your below code goes next
}
```

I'm not sure if this is anyhow faster due to the `TrimSpace` call; so I leave it up to you to decide if this approach makes the code any more readable.

vkhoroz (approved):
I think this is good, as it makes the user's life easier.

storage/api/api_storage.go
doanac: I think it helps readability. Claude wrote that loop and I did a double take the first time to understand the intent. Your idea makes the intent more clear.

doanac: fixed in force push

PR #150, opened 2026-04-29
feature: ability to manage local factory/group/device configs

vkhoroz:


server/ui/api/handlers_configs.go
```diff
@@ -47,3 +110,28 @@ One of them should be moved to the configs directory at '%s'.`,
...
+
+func configHistoryToJson(history []string) []json.RawMessage {
+	res := make([]json.RawMessage, len(history))
+	for i, cfg := range history {
+		res[i] = json.RawMessage(cfg)
+	}
+	return res
+}
+
+func validateConfigSet(c echo.Context) ([]byte, error) {
```
doanac: So this is basically like calling `Bind` but with the ability to get the original request bytes so you can avoid re-marshalling them?

server/ui/api/handlers_configs.go
```diff
@@ -47,3 +110,28 @@ One of them should be moved to the configs directory at '%s'.`,
...
+	res, err := io.ReadAll(body)
+	if err != nil {
+		return nil, EchoError(c, err, http.StatusBadRequest, "Failed to read request body")
+	}
+	dec := json.NewDecoder(bytes.NewReader(res))
+	dec.DisallowUnknownFields()
+	var configs ConfigFileSet
+	if err = dec.Decode(&configs); err != nil {
+		return nil, EchoError(c, err, http.StatusBadRequest, "Failed to parse request body")
+	}
```
doanac, line 270: For factory-wide config, we need to also reject z-50-fioctl stuff:

https://github.com/foundriesio/ota-lite/blob/main/ota_lite/api/factory.py#L470-L482

I know this isn't "factory wide", but for people that will do fleets with this - its to dangerous a setting

server/ui/api/handlers_configs.go
```diff
@@ -76,6 +76,67 @@ func (h *handlers) factoryConfigsPut(c echo.Context) error {
...
+// @Summary Save group configs, replacing current value
+// @Description Requires scopes: devices:read-update, updates:read-update
+// @Tags    Config
+// @Accept  json
+// @Param   name path string true "Group name"
+// @Param   data body ConfigFileSet true "Factory config files"
+// @Success 204
+// @Router  /configs/group/{name} [put]
+func (h *handlers) groupConfigsPut(c echo.Context) error {
+	group := c.Param("name")
```
doanac, line 133: We should probably have a regular expression to specify what makes a valid group name, right? e.g `validLabelValueRegex`

doanac: I also wonder if we should allow creation of a group config for a group that doesn't exist yet? I get that you need to handle it from the "import" logic. However, for this instance - what are the odds it was a mistaken typo versus an intentional operation? Or maybe there's a "--force" flag?

server/ui/api/handlers_configs.go
```diff
@@ -4,15 +4,78 @@
...
+
+const ConfigHistoryLimit = storage.ConfigHistoryLimit
+
+// @Summary Read latest factory configs
+// @Description Requires scopes: devices:read
+// @Tags    Config
+// @Produce json
+// @Success 200 {object} ConfigFileSet
+// @Router  /configs/factory [get]
+func (h *handlers) factoryConfigsGet(c echo.Context) error {
```
doanac, line 33: I'll see how the PR plays out, but my initial thought is having a both a single "latest" and a history GET is going to get repetitive and for not much benefit. 

What if we just had `/configs/factory?limit=1` to allow getting only the latest?

cli/subcommands/configs/set.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+var setCmd = &cobra.Command{
+	Use:   "set [ -g <group-name> | -d <device-uuid> ] <file1=content1> [ <file2=content2> ... ]",
+	Short: "Create a global, group, or device configs",
+	Long: `Create a global, group, or device configs.
+
+	By default, the command sets global configs.
+	If the --group or --device argument is provided, it sets group or device configs.
+	If a specified group does not exist - a command creates it.
```
doanac: Maybe "If a specified group does not exist, it will be created"?

cli/subcommands/configs/updates.go
```diff
@@ -0,0 +1,149 @@
...
+			cobra.CheckErr(sota.Delete("pacman.docker_apps"))
+		}
+		if sota.Has("pacman.compose_apps") {
+			cobra.CheckErr(sota.Delete("pacman.compose_apps"))
+		}
+	case ",":
+		apps = ""
+		fallthrough
+	default:
+		sota.Set("pacman.docker_apps", apps)
```
doanac: i think we can finally drop the legacy "docker_apps" setting - only compose_apps is used

doanac (commented):
It seems you are moving in the direction of de-emphasizing the idea of config history. I wonder if we should instead account for it in something like `satcli config show` but default to a limit of 1. I find myself looking at the history quite often to understand what fleet operations I've doing

server/ui/api/handlers_configs.go
vkhoroz: Exactly.
Configs map can be rather large, so avoiding one more serialization step is worth an optimization.

server/ui/api/handlers_configs.go
vkhoroz: I agree on group name validation.

As for exists/not exists, here are my thought...

We never create groups explicitly now - they are created implicitly when the device is assigned some label.
Now, I think a "safe" workflow would be to create a config for the group first, and then move devices into it.
That's why I decided to allow creating groups implicitly via config too.

As for a "typo" argument: there is a bigger danger if a user modifies existing config due to a typo (affects existing devices) than creating a new group via a typo (nothing gets affected). TBH I don't buy a typo danger unless we plan to become a kindergarten.

server/ui/api/handlers_configs.go
vkhoroz: A plain GET does more than "limit=1":
- It can signal that there is no config rather than returning an empty list.
  (That said, I'm not sure if I ever planned to use that).
- It returns a single item - less churn for a client side.

I may try thinking how to "reuse" here; but imho there is not much duplication here.

server/ui/api/handlers.go
```diff
@@ -26,6 +26,9 @@ func RegisterHandlers(e *echo.Echo, storage *storage.Storage, a auth.Provider) {
 
 	g.PUT("/configs", h.configsUpload, requireScope(users.ScopeDevicesRU|users.ScopeUpdatesRU),
 		gzipContentTypeAsContentEncoding, middleware.Decompress())
+	g.GET("/configs/factory", h.factoryConfigsGet, requireScope(users.ScopeDevicesR))
+	g.PUT("/configs/factory", h.factoryConfigsPut, requireScope(users.ScopeDevicesRU|users.ScopeUpdatesRU))
+	g.GET("/configs/factory/history", h.factoryConfigsHistory, requireScope(users.ScopeDevicesR))
```
vkhoroz, line 31: > It seems you are moving in the direction of de-emphasizing the idea of config history. I wonder if we should instead account for it in something like satcli config show but default to a limit of 1. I find myself looking at the history quite often to understand what fleet operations I've doing.

This is one of areas where I wanted a feedback _before_ deep diving into it.
A rationale: dg-satellite has less of config audit fields atm than FoundriesFactory (tbh, almost none of them).
A good history would give some rudimentary audit, and that requires some data storage layer changes...
Which might turn this PR into a marathon.

So, my plan was to:
1. Make an MVP without history/audit.
2. Discuss what's missing in this PR.
3. Add history/audit as a separate PR (if needed, and it looks like it is indeed needed).

server/ui/web/templates/configs_item.html
```diff
@@ -0,0 +1,37 @@
...
+                 <h3>
+                   File "{{ $name }}" content:
+                   <button onclick="this.closest('dialog').close()">x</button>
+                 </h3>
+                 <p>{{ $val.Value }}</p>
+                 <button autofocus onclick="this.closest('dialog').close()">Close</button>
+               </dialog>
+               {{ else }}
+               <i class="icontip lock-closed" aria-label="File content is encrypted"></i>
+               {{ end }}
```
vkhoroz, line 26: This PR does not add ability to upload configs via the Web UI, and I was wondering if we should do that.
If we should - we can allow only a plain content upload like FoundriesFactory, or a genuine file upload.
I'm inclined to support a genuine file upload... @doanac WDYT?

cli/subcommands/configs/set.go
```diff
@@ -0,0 +1,128 @@
...
+	"value": "root",
+	"on-changed": ["/usr/share/fioconfig/handlers/custom-handler", "/tmp/npmtok-changed"]
+	}
+	]
+	}
+	> EOF
+	satcli configs set --raw ./tmp.json
+
+	# satcli will read in tmp.json and upload it to the OTA server.
+	# Instead of using ./tmp.json, the command can take a "-" and will read the content from STDIN instead of a file.`,
```
vkhoroz, line 59: @doanac you might have noticed that there is no support for ECIES encryption in this PR.
A question: is this something we want to support for the satellite?
If yes - my plan was to add that as a separate PR, and it brings quite some dependencies.

server/ui/api/handlers.go
vkhoroz: Audit fields currently NOT present for configs are: `Reason, Author, AppliedAt`. Only `CreatedAt` is stored, but never used/exposed outside the gateway internals.
I think that for good history log we need all those fields, except (maybe) `AppliedAt` (and associated complexity).
@doanac what is your take?

server/ui/web/templates/configs_item.html
```diff
@@ -0,0 +1,37 @@
...
+         </tbody>
+      </table>
+{{ end }}
+
+{{ template "header" .}}
+    <section class="content-section">
+      <h2>{{.Title}}</h2>
+      {{ template "configs-item" .Configs }}
+    </section>
+{{ template "footer"}}
```
vkhoroz, line 39: This is how the UI currently looks like (in my dark widescreen theme on Linux).

Image

Image

Image

Image

cli/subcommands/configs/set.go
doanac: we'll need it, but not right this minute

server/ui/web/templates/configs_item.html
doanac: this is enough for now. I think the next step would be to make it easy to switch apps/tags.

server/ui/api/handlers.go
doanac: AppliedAt is important for a device config to know if/when a device has updated to the latest config. This is also why config history is important.

server/ui/web/templates/configs_item.html
```diff
@@ -0,0 +1,40 @@
...
+                 <h3>
+                   File "{{ $name }}" content:
+                   <button onclick="this.closest('dialog').close()">x</button>
+                 </h3>
+                 <p>{{ $val.Value }}</p>
+                 <button autofocus onclick="this.closest('dialog').close()">Close</button>
+               </dialog>
+               {{ else }}
+               <i class="icontip lock-closed" aria-label="File content is encrypted"></i>
+               {{ end }}
```
doanac, line 26: picocss already does some styling to do dialogs. You can make this:
```
              <dialog>
                 <article>
                   <header>
                     <button aria-label="Close" rel="prev" onclick="this.closest('dialog').close()"></button>
                     <h3>File "{{ $name }}" content</strong></h3>
                   </header>
                   <pre>{{ $val.Value }}</pre>
                  </article>
               </dialog>
```

and it will look arguably better and not require any css.

server/ui/web/templates/configs_item.html
```diff
@@ -0,0 +1,40 @@
...
+      </table>
+{{ end }}
+
+{{ template "header" .}}
+    <section class="content-section">
+      <h2>{{.Title}}</h2>
+      {{ template "configs-item" .Configs }}
+      {{ if .BackUrl }}
+      <button onclick="location.href='{{ .BackUrl }}';">Back</button>
+      {{ end }}
```
doanac: Let's leave out navigation stuff like this. We'll have to implement this for the whole project at some point and will want to do something consistent.

server/ui/web/templates/configs_list.html
```diff
@@ -0,0 +1,120 @@
...
+    <section class="content-section">
+      <h2>Groups</h2>
+      <table class="striped">
+        <thead>
+          <tr>
+            <th class="sortable">
+              <a class="groups-header" href="#" role="button"
+                 title="Sorted ascending, click for descending">UUID <span class="sort-active">▲</span></a>
+              <a class="groups-header hidden" href="#" role="button"
+                 title="Sorted descending, click for ascending">UUID <span class="sort-active">▼</span></a>
```
doanac, line 16: I'm not sure the role=button looks good. If you leave it out, you still see the sort icon and hovering gives you the mouse indicator.

server/ui/web/templates/configs_item.html
vkhoroz: image


It did not style well, as the "close" button gets above the header on the top-left side.
More critically, it always fits to text, that looks awful when the file lines are longer than the screen or the file is big in general. Actually, it was nearly dysfunctional for medium files on small screens.

So, I used your structure just for the semantic reader, but retained my styling which positions things properly.
The only change - I removed the "border: none" for buttons, as I suspect you did not like it.

vkhoroz: https://github.com/foundriesio/dg-satellite/pull/150/commits/4bf81dca481113a7d08cf8ca02b1407b003c1d75

server/ui/web/templates/configs_item.html
vkhoroz: Removed in force-push

server/ui/web/templates/configs_list.html
vkhoroz: The look-and-feel is exactly the same in both cases.
A `role="button"` simply tells the screen reader that this link behaves like a button i.e. does not redirect to another page.

cli/subcommands/configs/updates.go
vkhoroz: https://github.com/foundriesio/dg-satellite/pull/150/changes/9a06ce7e08219ce1fc3d3efb8fcb2fb817d20866

cli/subcommands/configs/set.go
vkhoroz: https://github.com/foundriesio/dg-satellite/pull/150/changes/c25c187ad8e1a860ff9c4913c6b0411508c3607b

server/ui/api/handlers_configs.go
vkhoroz: I was thinking about reusing code, but initial attempts add more code, so I just leave it as is.

server/ui/api/handlers_configs.go
vkhoroz: https://github.com/foundriesio/dg-satellite/pull/150/changes/8e5148057646d996470e86dfc8b35fa0eb41b4d8

For group name validation.

server/ui/api/handlers_configs.go
vkhoroz: https://github.com/foundriesio/dg-satellite/pull/150/changes/52e6ac17400a6a1157bc0228c9ee24a5e6e0d053

server/ui/api/handlers_configs.go
```diff
@@ -89,6 +90,10 @@ func (h *handlers) groupConfigsGet(c echo.Context) error {
 		return EchoError(c, err, http.StatusInternalServerError, "Failed to read group config history")
 	} else if len(history) > 0 {
 		return c.JSON(http.StatusOK, json.RawMessage(history[0]))
+	} else if knownGroups, err := h.storage.GetKnownDeviceGroupNames(); err != nil {
+		return EchoError(c, err, http.StatusInternalServerError, "Failed to get known group names")
+	} else if slices.Contains(knownGroups, group) {
+		return c.NoContent(http.StatusNoContent)
```
vkhoroz, line 96: I added a set of fixups (to be squashed later), which changes how the 204/404/200 statuses are handled for the non-existent device/group/config.
An idea is that for existing device/group we should not return the 404 status, even if the config does not exist: 204 or 200 status is more appropriate in this case.
I also fixed the 204 status code handling in Web UI and CLI as a part of this.

This affects multiple places:
- API return for known groups: https://github.com/foundriesio/dg-satellite/pull/150/changes/a1734a7822ac7489a8b4e71edc86b1c019c0e6fb;
- API return for existing devices: https://github.com/foundriesio/dg-satellite/pull/150/changes/8254789e5bc6c5b0d16970d7300994399597ceae
- CLI client handling of empty responses: https://github.com/foundriesio/dg-satellite/pull/150/changes/7dc8d701ab3d5b0dd6f49b7fe627d3e1bdcfbf1e;
- Web UI client handling of 204 status code: https://github.com/foundriesio/dg-satellite/pull/150/changes/25d260bff68ea5839bdc124661e368a2e468e58e;
- remove my "awkward patches" for handling 404 gracefully from WUI/CLI: https://github.com/foundriesio/dg-satellite/pull/150/changes/d52fbc8204175a39411f18262c94f91d8e7d6a19, https://github.com/foundriesio/dg-satellite/pull/150/changes/f7af8c274084ae77f168f70fa702d7d76fd850ca, https://github.com/foundriesio/dg-satellite/pull/150/changes/546a4d877e927d4dfbf462026c8b340c521e96cd.

server/ui/web/templates/configs_item.html
doanac: weird. might be something with dark vs light theme. we'll have milo eventually work through this, so no sense nit picking.

server/ui/web/templates/configs_list.html
doanac: there must be some dark vs light mode stuff happening. It looked different for me. However, your screen reader argument has enough merit for me.

vkhoroz: I finally managed to enable themes on my Linux Firefox (Ubuntu makes that hard), so I can style both light and dark themes now.
I've seen no problem with this header in any scheme.

But, I noticed a visual discrepancy in light scheme in a new icon for encrypted files, so I fixed that: https://github.com/foundriesio/dg-satellite/pull/150/commits/754b47f469fdcdb2d87731de891b3d432b7fc96f.

tbh I hate all those icons, but I didn't want to walk miles into searching for nice svg files with an appropriate license.
Maybe, we'll add font-awesome at some point, which would give us a much better visual nicety.

doanac (approved):

PR #151, opened 2026-04-30
Allow user to specify bind host

doanac:
Fixes issue #106

### DIFF
--- a/cmd/server/serve.go
+++ b/cmd/server/serve.go
@@ -20,8 +20,8 @@ import (
 type ServeCmd struct {
 	startedCb func(uiAddress, gatewayAddress string)
 
-	UiPort      uint16 `default:"8080"`
-	GatewayPort uint16 `default:"8443"`
+	UiAddr      string `default:":8080"`
+	GatewayAddr string `default:":8443"`
 }
 
 func (c *ServeCmd) Run(args CommonArgs) error {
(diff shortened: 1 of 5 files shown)


cmd/server/serve.go
```diff
@@ -33,11 +34,11 @@ func (c *ServeCmd) Run(args CommonArgs) error {
 	if err != nil {
 		return fmt.Errorf("failed to load database: %w", err)
 	}
-	uiServer, err := ui.NewServer(args.ctx, db, fs, c.UiPort)
+	uiServer, err := ui.NewServer(args.ctx, db, fs, fmt.Sprintf("%s:%d", c.BindHost, c.UiPort))
 	if err != nil {
 		return err
 	}
-	gtwServer, err := gateway.NewServer(args.ctx, db, fs, c.GatewayPort)
+	gtwServer, err := gateway.NewServer(args.ctx, db, fs, fmt.Sprintf("%s:%d", c.BindHost, c.GatewayPort))
```
vkhoroz: It's a shame that we need to join here, only to split inside a function.
But, it is also a trade-off... I see no ideal implementation.

One thing to keep in mind is that IPv6 addresses have some specifics. For example, look at the https://pkg.go.dev/net#JoinHostPort implementation. So, maybe, we should use `JoinHostPort` here.

cmd/server/serve.go
```diff
@@ -20,6 +20,7 @@ import (
 type ServeCmd struct {
 	startedCb func(uiAddress, gatewayAddress string)
 
+	BindHost    string `help:"Network interface address to bind to (e.g. 127.0.0.1 to bind only to loopback). Defaults to all interfaces."`
```
vkhoroz: I'd immediately make `UiPort, GatewayHost`, so that we don't have to rework it soon, when the first customer asks to separate networks.
I mean: a half of customers asking to configure a host would ask to configure separate hosts for "public" UI vs "private" device network.

vkhoroz (approved):

cmd/server/serve.go
doanac: interesting. fixed in force

cmd/server/serve.go
doanac: i agree - should have done that the first time. fixed in force push

PR #160, opened 2026-05-11
auth: Add csrf cookie for noauth

doanac:
We need a csrf cookie so noauth web sessions can make changes to things like device labels. This isn't really secure, but neither is noauth.

Fixes bug #159


vkhoroz (approved):
Thanks!

PR #161, opened 2026-05-14
Add support for understand a device's applied configuration

doanac:
@vkhoroz  - this is kind of a wild departure from how we do this in our product now, but I think this might have saved us a bunch of support issues over the years. It saves the exact config sent to a device so that you can understand what it has rather the interpolating it via a series of mental merges of config maps. It's slightly wasteful in terms of disk space, but I'm on the verge of being at peace with that considering the utility/observability this brings.

Consider this RFC. I just had Claude do it while you were reviewing my other PR

### DIFF
--- a/cli/api/configs.go
+++ b/cli/api/configs.go
@@ -6,12 +6,14 @@ package api
 import (
 	"io"
 
+	"github.com/foundriesio/dg-satellite/storage"
 	models "github.com/foundriesio/dg-satellite/storage/api"
 )
 
 type (
-	ConfigFile    = models.ConfigFile
-	ConfigFileSet = map[string]ConfigFile
+	ConfigFile     = models.ConfigFile
+	ConfigFileSet  = map[string]ConfigFile
+	AppliedConfigs = storage.AppliedConfigs
 )
 
 type ConfigsApi struct {
@@ -22,6 +24,8 @@ type SpecificConfigsApi struct {
 	api *Api
 	uri string
 }
+
+type DeviceConfigsApi struct{ SpecificConfigsApi }
 
 func (a *Api) Configs() ConfigsApi {
 	return ConfigsApi{api: a}
@@ -35,8 +39,8 @@ func (a ConfigsApi) Group(name string) SpecificConfigsApi {
 	return SpecificConfigsApi{api: a.api, uri: "/v1/configs/group/" + name}
 }
 
-func (a ConfigsApi) Device(uuid string) SpecificConfigsApi {
-	return SpecificConfigsApi{api: a.api, uri: "/v1/configs/device/" + uuid}
+func (a ConfigsApi) Device(uuid string) DeviceConfigsApi {
+	return DeviceConfigsApi{SpecificConfigsApi: SpecificConfigsApi{api: a.api, uri: "/v1/configs/device/" + uuid}}
 }
 
 func (a ConfigsApi) ListGroups() (names []string, err error) {
@@ -56,5 +60,16 @@ func (a SpecificConfigsApi) Get() (res ConfigFileSet, err error) {
 
 func (a SpecificConfigsApi) Put(configs ConfigFileSet) (err error) {
 	_, err = a.api.Put(a.uri, configs)
+	return
+}
+
+func (a DeviceConfigsApi) GetApplied() (result *AppliedConfigs, err error) {
+	var applied AppliedConfigs
+	if err = a.api.Get(a.uri+"/applied", &applied); err != nil {
+		return
+	}
+	if applied.AppliedAt != 0 {
+		result = &applied
+	}
 	return
 }
--- a/cli/subcommands/configs/applied.go
+++ b/cli/subcommands/configs/applied.go
@@ -1,0 +1,45 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+package configs
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/foundriesio/dg-satellite/cli/api"
+)
+
+var appliedCmd = &cobra.Command{
+	Use:   "applied <device-uuid>",
+	Short: "Show the applied configuration for a device",
+	Long: `Show the merged (factory + group + device) configuration that was most
+recently delivered to the device, along with the time it was applied.`,
+	Args: cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		a := api.CtxGetApi(cmd.Context()).Configs().Device(args[0])
+		getDeviceApplied(a)
+	},
+}
+
+func init() {
+	ConfigsCmd.AddCommand(appliedCmd)
+}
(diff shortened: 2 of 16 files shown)


storage/models.go
```diff
@@ -46,6 +46,13 @@ type ConfigFile struct {
 	OnChanged   []string `json:"OnChanged,omitempty"`
 }
 
+// AppliedConfigEnvelope wraps the merged config sent to a device along with
+// the Unix timestamp (seconds) at which it was delivered.
+type AppliedConfigEnvelope struct {
```
vkhoroz: IMHO the world `Envelope` is loaded here. Also, I deliberately used plural form `XyzConfigs` to distinguish from the dg-satellite config, so let's stick to it everywhere. Thus, an `AppliedConfigs` is just fine, or maybe `AppliedDeviceConfigs`.

storage/models.go
```diff
@@ -46,6 +46,13 @@ type ConfigFile struct {
...
+	AppliedAt int64                  `json:"applied_at"`
+	Config    map[string]*ConfigFile `json:"config"`
```
vkhoroz: We were using `AppliedAt` and `Files` in FoundriesFactory, so I guess it makes sense to copy that notation.
Just makes things easier to find/copy.

FYI I am also wrapping the configs in my work on history, and I'm using `Files` over there for the same thing.

server/ui/api/handlers_configs.go
```diff
@@ -168,6 +169,28 @@ func (h *handlers) deviceConfigsGet(c echo.Context) error {
...
+// @Description delivered to the device, along with the Unix timestamp when it was applied.
+// @Description Requires scopes: devices:read
+// @Tags    Config
+// @Produce json
+// @Param   uuid path string true "Device UUID"
+// @Success 200 {object} AppliedConfig
+// @Router  /configs/device/{uuid}/applied [get]
+func (h *handlers) deviceAppliedConfigGet(c echo.Context) error {
+	return h.handleDevice(c, func(device *Device) error {
+		envelope, err := h.storage.ReadAppliedConfig(device.Uuid)
```
vkhoroz: Can we agree to use `WriteAppliedConfigs/ReadAppliedConfigs` (preferred) or `SaveAppliedConfigs/LoadAppliedConfigs`?
I prefer the first option as that's how we structure all other `WriteXyz/ReadXyz` parts.

cli/api/configs.go
```diff
@@ -58,3 +60,14 @@ func (a SpecificConfigsApi) Put(configs ConfigFileSet) (err error) {
 	_, err = a.api.Put(a.uri, configs)
 	return
 }
+
+func (a ConfigsApi) DeviceApplied(uuid string) (result *AppliedConfigEnvelope, err error) {
```
vkhoroz: I think it'd be more aesthetic to declare `type DeviceConfigsApi struct { SpecificConfigsApi }` and then define `func (a DeviceConfigsApi) GetApplied() (result *AppliedConfigEnvelope, err error)`.
Why? `api.Configs().Device(uuid).GetApplied()` below is more in line with the spirit of this project.

cli/subcommands/configs/applied.go
```diff
@@ -0,0 +1,41 @@
...
+			fmt.Println("No configuration has been applied to this device yet.")
+			return
+		}
+		appliedAt := time.Unix(envelope.AppliedAt, 0)
+		fmt.Printf("Applied at: %s\n", appliedAt.Format(time.RFC1123))
+		cfg := make(api.ConfigFileSet, len(envelope.Config))
+		for k, v := range envelope.Config {
+			cfg[k] = *v
+		}
+		printConfigs(cfg)
```
vkhoroz: IIRC we've agreed to push the meat of command handlers into a separate function.
I'm not against other styles; just one style per project is easier to maintain.

server/ui/web/handlers_configs.go
```diff
@@ -5,11 +5,44 @@ package web
...
 import (
 	"fmt"
+	"time"
 
 	"github.com/foundriesio/dg-satellite/server/ui/api"
+	storage "github.com/foundriesio/dg-satellite/storage"
 	"github.com/labstack/echo/v4"
 )
 
+func (h handlers) configsDeviceApplied(c echo.Context) error {
```
vkhoroz: Can you rename this into `configsDeviceItemApplied` and more below `configsDeviceItem`?

server/ui/web/handlers_configs.go
```diff
@@ -5,11 +5,44 @@ package web
...
+	ctx := struct {
+		baseCtx
+		Configs   api.ConfigFileSet
+		AppliedAt string
+		BackUrl   string
+	}{
+		baseCtx:   h.baseCtx(c, fmt.Sprintf("Device \"%s\" Applied Config", uuid), "devices"),
+		Configs:   configs,
+		AppliedAt: appliedAt,
+		BackUrl:   "/devices/" + uuid,
```
vkhoroz: There is no `BackUrl` anymore as per your request.

vkhoroz (commented):
LGTM

My comments are fairly minimal; so I think we're close to merge this.

I have some thoughts how to extend this during my configs history work; that is out of scope for this PR.

server/ui/api/handlers_configs.go
```diff
@@ -168,6 +169,28 @@ func (h *handlers) deviceConfigsGet(c echo.Context) error {
...
+// @Summary Read the applied configuration last sent to a device
+// @Description Returns the merged config (factory + group + device) that was most recently
+// @Description delivered to the device, along with the Unix timestamp when it was applied.
+// @Description Requires scopes: devices:read
+// @Tags    Config
+// @Produce json
+// @Param   uuid path string true "Device UUID"
+// @Success 200 {object} AppliedConfigs
+// @Router  /configs/device/{uuid}/applied [get]
+func (h *handlers) deviceAppliedConfigGet(c echo.Context) error {
```
vkhoroz: Can this be `deviceAppliedConfigsGet` (plural, like everywhere), please?

cli/api/configs.go
```diff
@@ -23,6 +25,8 @@ type SpecificConfigsApi struct {
 	uri string
 }
 
+type DeviceConfigsApi struct{ SpecificConfigsApi }
```
vkhoroz, line 28: Hmm.... it is strange that both linter and formatter passed this.
There needs to be at least a space before  `{`.

server/ui/web/handlers_configs.go
```diff
@@ -5,11 +5,42 @@ package web
...
 import (
 	"fmt"
+	"time"
 
 	"github.com/foundriesio/dg-satellite/server/ui/api"
+	storage "github.com/foundriesio/dg-satellite/storage"
 	"github.com/labstack/echo/v4"
 )
 
+func (h handlers) configsDeviceItemApplied(c echo.Context) error {
```
vkhoroz: This would look more naturally at the bottom of this module.

vkhoroz (approved):

cli/api/configs.go
doanac: no - the linter does it this way for one-line statements like this

PR #163, opened 2026-05-19
db: Enable WAL and busy timeout

doanac:
These are common sqlite flags that can really improve scaling. Running device registration tests with Locust it started at:

5000 devices / spawn rate 20
----------------------------
 NO WAL
 * 2915 failures
 * 132 requests per second

 WITH WAL and BUSYTIMEOUT
 * 0 failures
 * 297 requests per second
 * 310ms average response time (99% - 690ms)

5000 devices / spawn rate 80
----------------------------
 WITH WAL and BUSYTIMEOUT
 * 0 failures
 * 380 requests per second
 * 670ms average response time (99% - 1900)

### DIFF
--- a/storage/db.go
+++ b/storage/db.go
@@ -36,10 +36,16 @@ func NewDb(dbfile string) (*DbHandle, error) {
 	if _, err := os.Stat(dbfile); err != nil {
 		newDb = errors.Is(err, os.ErrNotExist)
 	}
-	db, err := sql.Open("sqlite3", dbfile)
+	// _busy_timeout: retry for up to 5s instead of immediately returning SQLITE_BUSY.
+	// _journal_mode=WAL: allows concurrent readers while a write is in progress.
+	dsn := dbfile + "?_busy_timeout=5000&_journal_mode=WAL"
+	db, err := sql.Open("sqlite3", dsn)
 	if err != nil {
 		return nil, err
 	}
+	// A single writer connection prevents writer-vs-writer lock contention.
+	// Readers share the WAL file and are not blocked by this limit.
+	db.SetMaxOpenConns(1)
 	if newDb {
 		if err := createTables(db); err != nil {
 			return nil, err


storage/db.go
```diff
@@ -36,10 +36,16 @@ func NewDb(dbfile string) (*DbHandle, error) {
...
+	// _busy_timeout: retry for up to 5s instead of immediately returning SQLITE_BUSY.
+	// _journal_mode=WAL: allows concurrent readers while a write is in progress.
+	dsn := dbfile + "?_busy_timeout=5000&_journal_mode=WAL"
+	db, err := sql.Open("sqlite3", dsn)
 	if err != nil {
 		return nil, err
 	}
+	// A single writer connection prevents writer-vs-writer lock contention.
+	// Readers share the WAL file and are not blocked by this limit.
+	db.SetMaxOpenConns(1)
```
vkhoroz, line 48: One bottleneck I see is the limit of writers to just 1.
For the user facing API that's just fine, as satellite server is ought to be used by one customer only (limited number of users).
For the device facing API that might be a problem when a number of devices grows.
But, I guess, by the time a single customer crosses above e.g. 50'000 devices (where I suspect that might start to bother them), they'd find a way to migrate to e.g. PostgreSQL.  Alas, GitHub was running on a single-node MySQL instance until 2018 IIRC.

vkhoroz (approved):
I think this is the right way to go for now.

storage/db.go
doanac: sqlite only allows a single writer. Based on experiments as well as googling, this is the recommended way to run this. When I *don't* enable this line I requests/second fall from 230 to 120 and response times goes from 630ms to 700.

The bigger optimization is to have a read-only connection (this is safe because of WAL) and a write-only connection. It's actually fairly straightforward to do in our code, but I think we can wait to after 1.0.

vkhoroz: I didn't mean you did anything wrong - you did it the right way, as sqlite is by design synchronized for writes (i.e. only one writer at any given time is allowed whatever we do).
I just meant that this design limitation may become a problem in a distant future when a number of devices grows beyond some (big enough) threshold.

PR #165, opened 2026-05-26
feature: show configs history with audit fields

vkhoroz:


server/ui/web/templates/style.css
```diff
@@ -245,6 +245,99 @@ dialog.file-content p {
...
+.accordion-table-container {
+  display: grid;
+}
+
+.accordion-table-header,
+.accordion-table-body,
+.accordion-row-wrapper,
+.accordion-row-wrapper > details {
+  display: contents !important;
+}
```
doanac, line 262: I haven't looked too close, but picocss has an accordion view that may save you some effort: https://picocss.com/docs/accordion

doanac: i just did some light testing it works fine. At the end of this PR  - you should flag the "configs" CLI command to not be hidden.

Merge as soon as the test passes - we don't need to overthink this.

vkhoroz: The PicoCSS accordion is what I wanted to use.
A problem with it: we have a table and each row is an accordion. Means: cols inside the summary must align properly, and the article is a colspan; and that must be responsive.
So, Pico can't do that natively... I bet some heavier libraries like Bootstrap can; but we're not onto that yet.

vkhoroz: fwiw a part of styles here is specifically to disable what Pico does (those marked as important).

server/ui/web/templates/templates.go
```diff
@@ -29,7 +29,7 @@ func init() {
 		"contains": strings.Contains,
 		"deref": func(v any) any {
 			// Used to convert e.g. *bool to bool in config_item.html.
-			if val := reflect.ValueOf(v); val.Kind() == reflect.Ptr && !val.IsNil() {
+			if val := reflect.ValueOf(v); val.Kind() == reflect.Pointer && !val.IsNil() {
```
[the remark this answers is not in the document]
vkhoroz: It was golangci-lint who told to replace Ptr with Pointer in the newer Golang version (1.25).

cli/subcommands/configs/updates.go
```diff
@@ -65,22 +66,25 @@ func init() {
 	addSpecificFlags(updatesCmd)
 	updatesCmd.Flags().StringP("apps", "", "", "A comma-separated list of apps to run on the device.")
 	updatesCmd.Flags().StringP("tag", "", "", "A tag to follow on the device.")
+	updatesCmd.Flags().StringP("reason", "m",
+		"Override aktualizr-lite update configuration",
```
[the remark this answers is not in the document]
vkhoroz: It is used as a default value, not description.

cli/subcommands/configs/show.go
```diff
@@ -34,12 +35,18 @@ func init() {
...
 }
 
-func printConfigs(configs api.ConfigFileSet) {
-	fmt.Println("Files:")
-	for name, file := range configs {
+func printConfigsInfo(cfg api.ConfigFileSet) {
+	fmt.Printf("Reason: %s\n", cfg.Reason)
+	fmt.Printf("Created At: %s\n", time.Unix(cfg.CreatedAt, 0).UTC().Format(time.RFC3339))
+	fmt.Printf("Created By: %s\n", cfg.CreatedBy)
+}
```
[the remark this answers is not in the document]
vkhoroz: fixed in force-push

server/ui/api/handlers_configs.go
```diff
@@ -269,33 +283,76 @@ One of them should be moved to the configs directory at '%s'.`,
...
-		res[i] = json.RawMessage(cfg)
+		res[i] = configItemToJson(cfg)
 	}
 	return res
 }
 
-func validateConfigSet(c echo.Context, denySotaOverride bool) ([]byte, error) {
-	// We only need to validate config files, and return the original body (save on serialization)
+var (
+	// For a reason allow alphanum + basic puntcuation + space. Important: no newlines.
```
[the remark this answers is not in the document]
vkhoroz: fixed in force-push

cli/subcommands/configs/history.go
```diff
@@ -0,0 +1,53 @@
...
+	Use:   "history [ -g <group-name> | -d <device-uuid> ]",
+	Short: "Show a global, group, or device configs history",
+	Long: `Show a global, group, or device configs history.
+
+	By default, the command shows global configs history without files contents.
+	If the --group or --device argument is provided, it shows group or device configs history.
+	If --show-files argument is provided, it shows configs files contents for each history item.
+	Use the --limit argument to control how many history items are being shown.
+
+	Each configs history item is separated by a "----" line, with black lines above and below it.`,
```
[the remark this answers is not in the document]
vkhoroz: fixed in force-push

server/ui/web/templates/configs_item.html
```diff
@@ -31,14 +32,31 @@ <h3>File "{{ $name }}" content:</h3>
...
+      <button onclick="location.href=location.pathname+'/history'">History</button>
+    </section>
+{{ template "footer"}}
+
+{{ define "applied_configs_item.html" }}
+  {{ template "header" .}}
+    <section class="content-section">
+      <h2>{{.Title}}</h2>
+      {{ if gt .Configs.AppliedAt 0 }}
+      <p><b>Applied at:</b> {{ tsToString .AppliedAt }}</p>
```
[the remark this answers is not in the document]
vkhoroz: fixed in force-push

cli/subcommands/configs/log.go
```diff
@@ -0,0 +1,57 @@
...
+func showConfigsHistory(capi api.SpecificConfigsApi, limit int, showFiles bool) {
+	history, err := capi.GetHistory(limit, showFiles)
+	cobra.CheckErr(err)
+	if len(history) == 0 {
+		fmt.Println("No configuration has been created yet.")
+		return
+	}
+	for idx, cfg := range history {
+		if idx > 0 {
+			fmt.Print("\n----\n\n") // A linter would complain about a trailing newline in Println.
```
vkhoroz, line 50: Linter was complaining about Println with trailing "\n" here.
What's worse, even when that warning was disabled, unit tests were still complaining about this line.

fixed in force-push

server/ui/web/templates/configs_history.html
```diff
@@ -0,0 +1,46 @@
...
+              <summary data-index="{{ $idx }}">
+                <div>{{ tsToString $cfg.CreatedAt }}</div>
+                <div>{{ $cfg.CreatedBy }}</div>
+                <div>{{ $cfg.Reason }}</div>
+                <article onclick="event.preventDefault()">
+                  {{ if ge $.OpenIndex 0 }}
+                    {{ template "configs-item" $cfg }}
+                  {{ end }}
+                </article>
+              </summary>
```
vkhoroz, line 23: Here, I hate that the article is inside the summary, as the vanilla way is to place them side-by-side.
But, having them stacked is the only way how it works correctly (after I tried quite a few other solutions).
When side-by-side, article constantly fails to render the `span 3` properly into a grid system.

**TL;DR**
As per Gemini, the reason is that article is invisible at the page load when the CSS engine allocates slots for the grid system.
Hence, there are no slots allocated to it initially. Later, when it becomes visible - it is forced into the first available slot, and the span property is incorrectly processed (ignored).
Placing it inside the summary leverages a correct placement of that summary element.

I hardly believe into that storytelling, but I could not prove in practice that a correct layout is able to render correctly.

server/ui/web/templates/configs_history.html
```diff
@@ -0,0 +1,46 @@
+{{ template "header" .}}
+    <section class="content-section">
+      <h2>{{.Title}}</h2>
+      <div class="accordion-table-container configs-history">
```
vkhoroz, line 4: Some story just to capture the mental exercise I came through to make it work.

**TL;DR**

My initial version of this page was using tables. But, a table has some limitations which made it almost impossible to align the header with each row properly. A gargantuan style sheet was necessary for some basic visual nicety, 80% of which was removing native table styles.
Thus, going with the div based grid system seemed much more promising.

I bet that using divs instead of the details/summary/article would also render better; but that would require a JavaScript to implement an accordion. A few tweaks of how details/summary/article are composed inside the div+grid system gives us the best of two worlds.

doanac (approved):

server/ui/web/templates/configs_history.html
doanac: don't sweat it. This is the kind of thing copilot can deal with in the future.

PR #166, opened 2026-05-27
Include version information to users

doanac:
This allows users to understand and report the version of server they are working against.

### DIFF
--- a/server/ui/api/middleware.go
+++ b/server/ui/api/middleware.go
@@ -8,8 +8,11 @@ import (
 
 	"github.com/foundriesio/dg-satellite/auth"
 	"github.com/foundriesio/dg-satellite/storage/users"
+	"github.com/foundriesio/dg-satellite/version"
 	"github.com/labstack/echo/v4"
 )
+
+const versionHeader = "x-version"
 
 func requireScope(scope users.Scopes) echo.MiddlewareFunc {
 	return func(next echo.HandlerFunc) echo.HandlerFunc {
@@ -38,6 +41,8 @@ func authUser(provider auth.Provider) echo.MiddlewareFunc {
 			ctx = CtxWithLog(ctx, log)
 			ctx = CtxWithUser(ctx, user)
 			c.SetRequest(req.WithContext(ctx))
+
+			c.Response().Header().Set(versionHeader, version.Version)
 
 			return next(c)
 		}
--- a/server/ui/web/handlers.go
+++ b/server/ui/web/handlers.go
@@ -16,6 +16,7 @@ import (
 	"github.com/foundriesio/dg-satellite/server"
 	"github.com/foundriesio/dg-satellite/server/ui/web/templates"
 	"github.com/foundriesio/dg-satellite/storage/users"
+	"github.com/foundriesio/dg-satellite/version"
 )
 
 type handlers struct {
@@ -74,6 +75,7 @@ type baseCtx struct {
 	Title     string
 	NavItems  []navItem
 	CsrfToken string
+	Version   string
 }
 
 func (h handlers) baseCtx(c echo.Context, title, selected string) baseCtx {
@@ -86,6 +88,7 @@ func (h handlers) baseCtx(c echo.Context, title, selected string) baseCtx {
 		Title:     title,
 		NavItems:  h.genNavItems(selected),
 		CsrfToken: csrfToken,
+		Version:   version.Version,
 	}
 }
 
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 COMMIT?=$(shell git describe --tags HEAD)$(shell git diff --quiet || echo '+dirty')
 
 # Use linker flags to provide commit info
-LDFLAGS=-ldflags "-X=github.com/foundriesio/dg-satellite/cmd.Version=$(COMMIT)"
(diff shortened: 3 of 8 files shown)


server/ui/api/middleware.go
```diff
@@ -7,10 +7,13 @@ import (
 	"net/http"
 
 	"github.com/foundriesio/dg-satellite/auth"
+	"github.com/foundriesio/dg-satellite/cmd"
```
vkhoroz: It is weird to have an API import the `cmd` (especially having that `cmd` also imports an API indirectly i.e. we've got an almost cyclic imports).

If we want to return version of an API, I see 3 options:
- [Best] Extract the `cmd.Version` into a `github.com/foundriesio/dg-satellite/version` package, and reuse it everywhere.
- [Also good but risky] Have a version in `github.com/foundriesio/dg-satellite/version.go`; but we may still get into the cyclic imports via `main.go`.
- [Possible] Have a separate `version.go` file for the API, and let the release script fill it. IMHO this duplication can be avoided.

server/ui/api/middleware.go
```diff
@@ -7,10 +7,13 @@ import (
...
 	"github.com/foundriesio/dg-satellite/storage/users"
 	"github.com/labstack/echo/v4"
 )
 
+const versionHeader = "x-dg-sat-version"
```
vkhoroz: There is an _almost_ standard `X-Version` header for this.

cli/subcommands/version/cmd.go
```diff
@@ -0,0 +1,57 @@
...
+	cobra.CheckErr(err)
+
+	resp, err := client.Client.Do(req)
+	cobra.CheckErr(err)
+	defer func() {
+		_ = resp.Body.Close()
+	}()
+
+	version := resp.Header.Get(serverVersionHeader)
+	fmt.Println("server api version:", version)
```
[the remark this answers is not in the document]
vkhoroz: I guess Copilot is technically right here.

IIUC the GET request to `/v1` would return the 404 error, which still contains a version header. So, we ignore the error and use the header. That's fine, but not ideal.

Why don't we have a non-authorized `/v1/version` endpoint which returns a version as a plain string?
A version header is nice for debugging, but barely convenient otherwise.

cli/subcommands/version/cmd.go
```diff
@@ -5,15 +5,53 @@ package version
...
+	client := api.NewClient(*ctx)
+	url := strings.TrimRight(ctx.URL, "/") + "/v1"
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	cobra.CheckErr(err)
+
+	resp, err := client.Client.Do(req)
+	cobra.CheckErr(err)
+	defer func() {
+		_ = resp.Body.Close()
+	}()
```
vkhoroz: Why can't we use `api.GetWithHeaders`?

cli/subcommands/version/cmd.go
```diff
@@ -5,15 +5,53 @@ package version
...
+func init() {
+	VersionCmd.Flags().Bool("server", false, "Also query and print the server API version")
+}
+
+func serverAPIVersion(configPath, contextName string) {
+	cfg, err := config.LoadConfig(configPath)
+	cobra.CheckErr(err)
+
+	ctx, err := cfg.GetContext(contextName)
+	cobra.CheckErr(err)
```
vkhoroz: We don't seem to bother with config loading in other commands. Why is this one special?
In other words, how to make it behave _as usual_?

server/ui/web/handlers.go
```diff
@@ -67,10 +68,11 @@ func RegisterHandlers(e *echo.Echo, storage *users.Storage, authProvider auth.Pr
...
 type baseCtx struct {
-	User      *users.User
-	Title     string
-	NavItems  []navItem
-	CsrfToken string
+	User          *users.User
+	Title         string
+	NavItems      []navItem
+	CsrfToken     string
+	ServerVersion string
```
vkhoroz: Just `Version` would work as fine and cause less churn.

vkhoroz (commented):
The idea to show version is nice, and the UI looks pretty good.

I only have concerns about the version module itself, and the satcli boilerplate.

cli/subcommands/version/cmd.go
doanac: a user may not have a config when they want to run the version command. if the `login` fails and they want to report a bug, they need to be able to tell us the version of the binary.

vkhoroz: I think I get it now:
- `satcli version` does not load config.
- `satcli version --server` needs to load config.

That's why you copied that part of the code.

IMHO, we may do better here, as this `serverAPIVersion` is 100% the same as `rootCmd.PersistentPreRunE` when it comes to the API client initialization. That already deserves a common routine.

tbh, I think that a root command of `github.com/foundriesio/dg-satellite/cmd/cli/main.go` can be moved into `github.com/foundriesio/dg-satellite/cli/main.go`. Then, your `serverAPIVersion` could call `cli.RootCmd.PersistentPreRunE(cmd, []string{})` to initialize the API client, and the `cmd/cli/main.go` would call `cli.RootCmd.Execute()`. I'm just fantasizing.

vkhoroz (approved):
I only suggest to use an `X-Version` header, and then feel free to merge.

PR #171, opened 2026-06-05
web: ability to manage global/group/device configs from UI

vkhoroz:


server/ui/web/handlers_configs.go
```diff
@@ -156,3 +194,43 @@ func (h handlers) configsDeviceItemHistory(c echo.Context) error {
...
+		return EchoError(c, err, http.StatusBadRequest, "Could not parse request")
+	}
+	configs.Reason = patch.Reason
+	configs.Files[patch.FileName] = api.ConfigFile{
+		// Split by newline and skip white-space only lines.
+		OnChanged:   strings.FieldsFunc(patch.OnChanged, func(r rune) bool { return r == '\n' }),
+		Unencrypted: true,
+		Value:       patch.Value,
+	}
+	fmt.Printf("AAAAAAAAAAAAA %v %v\n", patch.OnChanged, configs.Files[patch.FileName].OnChanged)
```
doanac: left over debug

doanac: I'm guessing there's good reason, but it feels like you could almost call the REST API directly from the javascript handler and avoid this extra patch logic here?

server/ui/web/templates/configs_item_edit.html
```diff
@@ -0,0 +1,170 @@
...
+              el.style.width = el.scrollWidth + 'px';
+            });
+          }
+        }
+
+        if (!CSS.supports('field-sizing', 'content')) {
+          tx.forEach(el => el.addEventListener('input', autoResizeTextareas));
+        }
+      </script>
+{{ end }}
```
doanac, line 184: This is rendering a little weird for me:

Image

Also - let's drop the "onchangehandler" - we elected to *not* show that in our app.fio because people will misunderstand it. You can't put arbitrary things there - it must be a handler supported by your image.

server/ui/web/handlers_configs.go
```diff
@@ -156,3 +194,43 @@ func (h handlers) configsDeviceItemHistory(c echo.Context) error {
...
+	return nil
```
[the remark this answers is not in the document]
vkhoroz: fixed in force push.
As for the validation/sanitation - it is not needed for generic text.

server/ui/web/templates/configs_item_edit.html
```diff
@@ -0,0 +1,170 @@
...
+              [...document.querySelectorAll('dialog.file-content header > h3 > span')]
+              .some(el => el.textContent.trim() === name)
+            ) {
+              if ( !confirm(`A file with name "${name}" already exists. Do you want to overwrite it?`) ) {
+                return
+              }
+            }
+          }
+          saveConfigFile();
+        });
```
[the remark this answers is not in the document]
vkhoroz: fixed nil dereference in force-push.

There is no submit button intentionally to prevent an accidental form submit on <Enter>.
I verified that it works as intended.

server/ui/web/templates/configs_item_edit.html
```diff
@@ -0,0 +1,170 @@
...
+        </article>
+      </dialog>
+      <script type="module" id="EditConfigsFile">
+        export const config = {
+          'apiUrl': location.pathname, // PATCH usually sits at the same URL as the page itself.
+          'addButtonName': 'AddFile',
+          'editButtonName': 'EditFile',
+        }
+        /* export/import of script tags does not work in FireFox, so use the window object for that */
+        window.EditConfigsFileConfig = config;
```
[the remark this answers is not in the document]
vkhoroz: fixed in force push.

server/ui/web/templates/configs_item_edit.html
```diff
@@ -0,0 +1,170 @@
...
+          dialog.dataset.mode = "add";
+          dialogFileName.readOnly = false;
+          dialogFileName.value = "";
+          dialogOnChanged.value = "";
+          dialogReason.value = "";
+          dialogValue.value = "";
+          dialog.showModal();
+          /* field-sizing: content; for FireFox and Safari. */
+          autoResizeTextareas();
+        }
```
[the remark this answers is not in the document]
vkhoroz: fixed in force push.

server/ui/web/templates/configs_item_edit.html
```diff
@@ -0,0 +1,170 @@
...
+          ev.preventDefault();
+          const src = ev.target.closest('dialog')
+          dialog.dataset.mode = "edit";
+          dialogFileName.readOnly = true;
+          copyConfigsFileData(src);
+          src.close();
+          dialog.showModal();
+          /* field-sizing: content; for FireFox and Safari. */
+          autoResizeTextareas();
+        }
```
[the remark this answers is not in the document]
vkhoroz: fixed in force push.

server/ui/web/handlers.go
```diff
@@ -44,10 +44,13 @@ func RegisterHandlers(e *echo.Echo, storage *users.Storage, authProvider auth.Pr
 	e.GET("/auth/logout", h.authLogout, h.requireSession)
 	e.GET("/configs", h.configsList, h.requireSession, h.requireScope(users.ScopeDevicesR))
 	e.GET("/configs/device/:uuid", h.configsDeviceItem, h.requireSession, h.requireScope(users.ScopeDevicesR))
+	e.PATCH("/configs/device/:uuid", h.configsDeviceItemPatch, h.requireSession, h.requireScope(users.ScopeDevicesRU), auth.CsrfCheck)
 	e.GET("/configs/device/:uuid/applied", h.configsDeviceItemApplied, h.requireSession, h.requireScope(users.ScopeDevicesR))
 	e.GET("/configs/device/:uuid/history", h.configsDeviceItemHistory, h.requireSession, h.requireScope(users.ScopeDevicesR))
+	e.PATCH("/configs/global", h.configsGlobalPatch, h.requireSession, h.requireScope(users.ScopeDevicesRU), auth.CsrfCheck)
 	e.GET("/configs/global/history", h.configsGlobalHistory, h.requireSession, h.requireScope(users.ScopeDevicesR))
 	e.GET("/configs/group/:name", h.configsGroupItem, h.requireSession, h.requireScope(users.ScopeDevicesR))
+	e.PATCH("/configs/group/:name", h.configsGroupItemPatch, h.requireSession, h.requireScope(users.ScopeDevicesRU), auth.CsrfCheck)
```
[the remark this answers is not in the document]
vkhoroz: fixed in force push.

server/ui/web/handlers_configs.go
```diff
@@ -156,3 +194,43 @@ func (h handlers) configsDeviceItemHistory(c echo.Context) error {
...
+	}
+	if err := putJson(c.Request().Context(), "/v1/configs/device/"+uuid, configs, nil); err != nil {
+		return h.handleUnexpected(c, err)
+	}
+	return c.NoContent(http.StatusOK)
+}
+
+func (h handlers) configsEditable(c echo.Context) bool {
+	return CtxGetSession(c.Request().Context()).User.AllowedScopes.Has(users.ScopeDevicesRU)
+}
```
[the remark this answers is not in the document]
vkhoroz: fixed in force push.

server/ui/web/handlers_configs.go
vkhoroz: fixed debug line in force push.

I had 3 options:
- Write JavaScript to use generic API's PUT handler - too complex.
- Add generic API PATCH handler and write some custom JavaScript to convert form data to a map of config files - easy, but still needs JavaScript for data conversion.
- Add a specific Web PATCH handler which understands a plain form submit (and uses a generic API PUT) - **_that's what I did_**.

server/ui/web/templates/configs_item_edit.html
vkhoroz: 1. As for rendering - it's interesting what browser are you using. MDN says that Chrome must render `field-sizing: content` accurately, but I had no chance to test it properly. I suspect it does not. So, I changed to use JavaScript (not CSS) based resizing for all browsers, as I've tested it thoroughly, while `field-sizing` is quite new. **_Fixed in force-push_**.

2. As for OnChanged - if we remove it - we'd need a custom code to support the `z-50-fioctl.toml` in the UI. That's something I wanted to avoid, as otherwise a user would only be able to upload that file from the CLI. So, instead of disallowing to modify this property, I added a description about where a handler must be placed on the device.

server/ui/web/templates/configs_item.html
```diff
@@ -1,48 +1,112 @@
...
-               <i class="icontip ocr-fork" aria-label="On-change triggers:{{ range $val.OnChanged }}&#xa;{{.}}{{ end }}"></i>
+               {{ if gt (len .OnChanged) 0 }}
+               <i class="icontip ocr-fork" aria-label="On-change triggers:{{ range .OnChanged }}&#xa;{{.}}{{ end }}"></i>
                {{ end }}
-               {{ if $val.Unencrypted }}
-               <i class="icontip floppy" aria-label="View file content"
-		  onclick="event.preventDefault();this.nextElementSibling.showModal()"></i>
+               {{ if .Unencrypted }}
+               <i class="icontip gear" aria-label="View{{ if $.CanEdit }} or edit{{ end }} file content"
+                  onclick="event.preventDefault();this.nextElementSibling.showModal()"></i>
```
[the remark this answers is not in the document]
vkhoroz: nixed

server/ui/web/templates/configs_item.html
```diff
@@ -1,48 +1,112 @@
...
+                     {{ end }}
+                     <button autofocus onclick="this.closest('dialog').close()">Close</button>
+                   </footer>
                  </article>
                </dialog>
                {{ else }}
                <i class="icontip lock-closed" aria-label="File content is encrypted"></i>
                {{ end }}
+               {{ if $.CanEdit }}
+               <i class="icontip trashbin" aria-label="Delete file"></i>
```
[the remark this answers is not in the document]
vkhoroz: nixed

server/ui/web/templates/configs_item_edit.html
```diff
@@ -0,0 +1,184 @@
...
+              </fieldset>
+            </form>
+          </section>
+          <footer>
+            <button name="save">Save changes</button>
+            <button onclick="this.closest('dialog').close()">Cancel</button>
+          </footer>
+        </article>
+      </dialog>
+      <script type="module" id="EditConfigsFile">
```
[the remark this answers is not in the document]
vkhoroz: fixed in force-push

server/ui/web/http.go
```diff
@@ -37,12 +62,15 @@ func getJsonWithHeaders(ctx context.Context, resource string, result any) (http.
 
 	switch resp.StatusCode {
 	case http.StatusOK:
+		if result == nil {
+			// Caller requested to skip result data
+			return resp.Header, nil
+		}
 		return resp.Header, json.NewDecoder(resp.Body).Decode(result)
```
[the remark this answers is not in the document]
vkhoroz: fixed in force-push

server/ui/web/handlers_configs.go
```diff
@@ -156,3 +181,69 @@ func (h handlers) configsDeviceItemHistory(c echo.Context) error {
...
+	var patch struct {
+		Reason    string
+		FileName  string
+		OnChanged string
+		Value     string
+	}
+	if err := c.Bind(&patch); err != nil {
+		return EchoError(c, err, http.StatusBadRequest, "Could not parse request")
+	}
+	var configs api.ConfigFileSet
```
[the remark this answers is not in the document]
vkhoroz: fixed in force-push, without trimming (file names must be used verbatim).

server/ui/web/templates/style.css
```diff
@@ -187,6 +187,13 @@ i.ocr-fork::before {
   line-height: 1rem;
 }
 
+i.trashbin::before {
+  content: "\1F5D1\FE0E";
+  display: inline-block;
+  transform: translateY(-0.15rem);
+  font-style: normal;
+}
+
```
doanac: we already have a `i.trash` icon defined a few lines above this.

server/ui/web/templates/configs_item.html
```diff
@@ -1,48 +1,112 @@
...
-           {{ range $name, $val := . }}
+           {{ range $name, $_ := . }}
            <tr>
-             <td style="min-width: 20em;">{{ $name }}</td>
+             <td style="min-width: 20em;" name="FileName">{{ $name }}</td>
              <td style="min-width: 5em; text-align: right;">
-               {{ if gt (len $val.OnChanged) 0 }}
-               <i class="icontip ocr-fork" aria-label="On-change triggers:{{ range $val.OnChanged }}&#xa;{{.}}{{ end }}"></i>
+               {{ if gt (len .OnChanged) 0 }}
+               <i class="icontip ocr-fork" aria-label="On-change triggers:{{ range .OnChanged }}&#xa;{{.}}{{ end }}"></i>
```
doanac, line 11: "triggers" is confusing (we sometimes call remote actions "triggers"). Call this "On-change handlers"

server/ui/web/templates/configs_item_edit.html
doanac: I get it. This may be what held up progress for us ever adding this to our own web UI. I think there's a 30% chance we'll regret adding this feature like this. What if we show the on-change handler but don't allow editing it?

doanac (commented):
This is kind of nitpicking, but its a pain to go through later: Fields like "Created At" should only capitalize the first word. e.g "Created at". I won't debate this on grounds of grammar but of consistency with what we done elsewhere in the UI as well as what I copied from, app.fio.

server/ui/web/templates/configs_item_edit.html
vkhoroz: I usually prefer to make things easier to configure, not harder.
If we disable it here - it's still enabled in the CLI via a `--raw` option.

If you're afraid about users rushing to configure handlers "because they can", then what if we hide this textarea behind the checkbox (unchecked by default, hiding that textarea)?

But, in any case when this setting is "not-so-trivial-to-access" - we have to special-case the `z-50-fioctl.toml` in the UI (the same way we do it in the CLI). Or, do you think we need to special-case it anyway - that's not trivial at all, as it is quite dumb to modify `OnChanged` on-the-fly as user types a file name. A rather creative way might be to have a separate `Add Updates Config` button for that.

vkhoroz: TBH this discussion is entering a rabbit hole.
I strove to give a user smth simple to implement and generic enough to cover all use cases; and then add special cases or extend/restrain later when there is some practical feedback.

server/ui/web/templates/configs_item.html
vkhoroz: IIRC I copied this from the app.foundries.io; but I'm fine with what you propose here.

server/ui/web/templates/configs_item_edit.html
doanac: Looking at how our current webui is that Milo/Eric did was interesting. You can add/remove files - you can't edit them. That makes some sense considering files can be encrypted and/or have on-change handlers.

Or just merge this and see what happens.

server/ui/web/templates/configs_item.html
doanac: let's change it - the web ui is misleading

server/ui/web/templates/configs_item_edit.html
vkhoroz: You know what I plan to do as a person with an author bias. ;)

Anyway, I think any approach has its pros and cons, so I'd rather merge and see than sink in a discussion.
Somebody can change it in 3 weeks from now, that's life... or it can stay for 30 years.

fwiw encrypted files are not editable atm, they can be only removed.
If we ever add ECIES to the UI, I could see an "edit" for encrypted files via remove/add sequence, although nothing stops us from making an "edit" button which combines these two operations into one.

doanac (approved):
k - merge this. i'll then do a rename of the project and start updating all the import paths.

server/ui/web/templates/style.css
vkhoroz: Makes sense.

These image-based icons are a bit less flexible in styling than font-based I used here, but I found a way to re-use it.

server/ui/web/templates/configs_item.html
vkhoroz: fixed in force-push

PR #183, opened 2026-06-15
Drop the notion of "ci" and "production"  Updates and Devices

doanac:
This concept is really just needed for our multi-tenant SAAS backend. In the context of this project, its causing unneeded complexity. There is no easy way to cleanly break up the commits. I've tried to structure changes in some logical order but these changes don't produce a bisectable set of commits.

One important note: This is not a backward compatible change. People will existing deployments like myself will need to update their backend storage to deal with this.

Fixes: https://github.com/foundriesio/update-server/issues/181


server/ui/web/templates/device.html
```diff
@@ -80,20 +88,6 @@ <h2>{{.Title}}</h2>
...
-              <ul>
-              {{/* A name and a group labels (if present) are always the first ones */}}
-              {{ if .Device.Labels.name }}<li>name: {{.Device.Labels.name}}</li>{{ end }}
-              {{ if .Device.Labels.group }}<li>group: {{.Device.Labels.group}}</li>{{ end }}
-              {{ range $key, $value := .Device.Labels }}
-                {{if and (ne $key "name") (ne $key "group")}}<li>{{$key}}: {{$value}}</li>{{end}}
-              {{ end }}
-              </ul>
-            </dd>
-        </div>
```
vkhoroz, line 96: What is the value of moving the Update and Labels sections around?

server/gateway/handlers_test.go
```diff
@@ -577,12 +561,12 @@ func TestEvents(t *testing.T) {
 func TestTufMeta(t *testing.T) {
 	tcCi42 := NewTestClient(t)
 	tcCi137 := NewTestClient(t)
-	tcProd42 := NewProdTestClient(t)
-	tcProd137 := NewProdTestClient(t)
+	tcCi42b := NewTestClient(t)
+	tcCi137b := NewTestClient(t)
```
vkhoroz: If we remove Ci vs Prod distinction - it makes sense to rename Ci too, e.g. `tc42, tc137` etc.

IIRC, the whole idea of `tcProdXyz` devices was to test isProd. So, their tests can be safely removed.

storage/gateway/storage.go
```diff
@@ -129,11 +128,7 @@ func (d Device) ProcessEvents(events []storage.DeviceUpdateEvent) error {
 			if err != nil {
 				return err
 			}
-			fs := d.storage.fs.Updates.Ci.Logs
-			if d.IsProd {
-				fs = d.storage.fs.Updates.Prod.Logs
-			}
-			if err = fs.AppendFile(d.Tag, d.UpdateName, storage.LogRolloutsFile, string(bytes)+"\n"); err != nil {
+			if err = d.storage.fs.Updates.Ci.Logs.AppendFile(d.Tag, d.UpdateName, storage.LogRolloutsFile, string(bytes)+"\n"); err != nil {
```
vkhoroz: This line seems too long by any measure.
```suggestion
			if err = d.storage.fs.Updates.Ci.Logs.AppendFile(
				d.Tag, d.UpdateName, storage.LogRolloutsFile, string(bytes)+"\n"); err != nil {
```

vkhoroz (approved):
LGTM

I think this is the right way to go, if we're no longer going to support the integration with our SaaS platform.
Feel free to merge once the tests pass and you're happy with the result.

server/ui/web/templates/device.html
doanac: We are creating a grid with these divs:
image

By removing the "production" column - we wind up with a row of 3 items, a row of 4, and a row of 2. It looks weird. This change balances the columns to look like:

image

server/gateway/handlers_test.go
doanac: fixed in force push

storage/gateway/storage.go
doanac: fixed in force push

PR #184, opened 2026-06-15
Improvements to Updates API

doanac:
These enhancements were requested by the team during our last sprint:

 * include when the update was created
 * sort by that date
 * show who created the update

### DIFF
--- a/cli/subcommands/updates/list.go
+++ b/cli/subcommands/updates/list.go
@@ -4,6 +4,8 @@
 package updates
 
 import (
+	"time"
+
 	"github.com/foundriesio/update-server/cli/api"
 	"github.com/foundriesio/update-server/cli/subcommands"
 	"github.com/spf13/cobra"
@@ -27,11 +29,11 @@ func listUpdates(api *api.Api) {
 	allUpdates, err := api.Updates().List()
 	cobra.CheckErr(err)
 
-	t := subcommands.NewTableWriter([]string{"TAG", "NAME"})
+	t := subcommands.NewTableWriter([]string{"TAG", "NAME", "UPLOADED AT", "UPLOADED BY"})
 
-	for tag, names := range allUpdates {
-		for _, name := range names {
-			t.AddRow(tag, name)
+	for tag, updates := range allUpdates {
+		for _, update := range updates {
+			t.AddRow(tag, update.Name, time.Unix(update.UploadedAt, 0).Format(time.RFC3339), update.UploadedBy)
 		}
 	}
 
--- a/server/ui/api/handlers_test.go
+++ b/server/ui/api/handlers_test.go
@@ -608,17 +608,23 @@ func TestApiUpdateList(t *testing.T) {
 		return strings.TrimSpace(string(data))
 	}
 
+	require.Nil(t, tc.api.InsertUpdate("tag1", "update1"))
 	require.Nil(t, tc.fs.Updates.Rollouts.WriteFile("tag1", "update1", "rollout1", "foo"))
+
+	require.Nil(t, tc.api.InsertUpdate("tag1", "update2"))
 	require.Nil(t, tc.fs.Updates.Rollouts.WriteFile("tag1", "update2", "rollout1", "foo"))
+
+	require.Nil(t, tc.api.InsertUpdate("tag2", "update1"))
 	require.Nil(t, tc.fs.Updates.Rollouts.WriteFile("tag2", "update1", "rollout1", "foo"))
 	require.Nil(t, tc.fs.Updates.Rollouts.WriteFile("tag2", "update3", "rollout1", "foo"))
 
 	data := tc.GET("/updates", 200)
-	assert.Equal(t, `{"tag1":["update1","update2"],"tag2":["update1","update3"]}`, s(data))
+	assert.Equal(t, `{"tag1":["update1","update2"],"tag2":["update1"]}`, s(data))
+
 	data = tc.GET("/updates/tag1", 200)
 	assert.Equal(t, `{"tag1":["update1","update2"]}`, s(data))
 	data = tc.GET("/updates/tag2", 200)
-	assert.Equal(t, `{"tag2":["update1","update3"]}`, s(data))
+	assert.Equal(t, `{"tag2":["update1"]}`, s(data))
 	data = tc.GET("/updates/tag4", 200) // tag not exists
 	assert.Equal(t, "{}", s(data))
 
@@ -702,7 +708,9 @@ func TestApiRolloutPut(t *testing.T) {
 	tc.PUT("/updates/tag/update/rollouts/rocks", 400, "{}")
 
 	require.Nil(t, tc.fs.Updates.Ostree.WriteFile("tag1", "update1", "foo", "bar"))
+	require.Nil(t, tc.api.InsertUpdate("tag1", "update1"))
 	require.Nil(t, tc.fs.Updates.Ostree.WriteFile("tag2", "update2", "foo", "bar"))
+	require.Nil(t, tc.api.InsertUpdate("tag2", "update2"))
 	d, err := tc.gw.DeviceCreate("ci1", "pubkey1")
 	require.Nil(t, err)
 	require.Nil(t, d.CheckIn("", "tag1", "", ""))
@@ -1390,6 +1398,10 @@ func TestApiUpdateCreate(t *testing.T) {
 	appData, err := os.ReadFile(filepath.Join(updatesDir, "main", "v2.0", "apps", "myapp.json"))
 	require.NoError(t, err)
 	assert.Equal(t, `{"name":"myapp"}`, string(appData))
+
+	// Duplicate tag and name
+	tc.POST("/updates/main/v2.0", 409, bytes.NewReader(appsTar.Bytes()),
+		"Content-Type", "application/x-tar")
 
 	// Valid tar with tuf + ostree_repo + apps (both present)
 	bothTar := tarBuffer(t, map[string]string{
--- a/server/ui/api/handlers_test.go
+++ b/server/ui/api/handlers_test.go
@@ -619,13 +619,13 @@ func TestApiUpdateList(t *testing.T) {
 		return res
 	}
 
-	require.Nil(t, tc.api.InsertUpdate("tag1", "update1"))
+	require.Nil(t, tc.api.InsertUpdate("tag1", "update1", "user1"))
 	require.Nil(t, tc.fs.Updates.Rollouts.WriteFile("tag1", "update1", "rollout1", "foo"))
 
-	require.Nil(t, tc.api.InsertUpdate("tag1", "update2"))
+	require.Nil(t, tc.api.InsertUpdate("tag1", "update2", "user1"))
 	require.Nil(t, tc.fs.Updates.Rollouts.WriteFile("tag1", "update2", "rollout1", "foo"))
 
-	require.Nil(t, tc.api.InsertUpdate("tag2", "update1"))
+	require.Nil(t, tc.api.InsertUpdate("tag2", "update1", "user1"))
 	require.Nil(t, tc.fs.Updates.Rollouts.WriteFile("tag2", "update1", "rollout1", "foo"))
 	require.Nil(t, tc.fs.Updates.Rollouts.WriteFile("tag2", "update3", "rollout1", "foo"))
 
@@ -719,9 +719,9 @@ func TestApiRolloutPut(t *testing.T) {
 	tc.PUT("/updates/tag/update/rollouts/rocks", 400, "{}")
 
 	require.Nil(t, tc.fs.Updates.Ostree.WriteFile("tag1", "update1", "foo", "bar"))
-	require.Nil(t, tc.api.InsertUpdate("tag1", "update1"))
+	require.Nil(t, tc.api.InsertUpdate("tag1", "update1", "user1"))
 	require.Nil(t, tc.fs.Updates.Ostree.WriteFile("tag2", "update2", "foo", "bar"))
-	require.Nil(t, tc.api.InsertUpdate("tag2", "update2"))
+	require.Nil(t, tc.api.InsertUpdate("tag2", "update2", "user1"))
 	d, err := tc.gw.DeviceCreate("ci1", "pubkey1")
 	require.Nil(t, err)
 	require.Nil(t, d.CheckIn("", "tag1", "", ""))
--- a/server/ui/api/handlers_test.go
+++ b/server/ui/api/handlers_test.go
@@ -604,8 +604,19 @@ func TestApiUpdateList(t *testing.T) {
 	tc.GET("/updates/tag", 403)
 	tc.u.AllowedScopes = users.ScopeUpdatesR
 
-	s := func(data []byte) string {
-		return strings.TrimSpace(string(data))
+	updateNames := func(data []byte) map[string][]string {
+		var updates map[string][]apiStorage.Update
+		require.Nil(t, json.Unmarshal(data, &updates))
+		res := make(map[string][]string, len(updates))
+		for tag, upds := range updates {
+			names := make([]string, len(upds))
+			for i, u := range upds {
+				require.NotZero(t, u.UploadedAt)
+				names[i] = u.Name
+			}
+			res[tag] = names
+		}
+		return res
 	}
 
 	require.Nil(t, tc.api.InsertUpdate("tag1", "update1"))
@@ -619,14 +630,14 @@ func TestApiUpdateList(t *testing.T) {
 	require.Nil(t, tc.fs.Updates.Rollouts.WriteFile("tag2", "update3", "rollout1", "foo"))
 
 	data := tc.GET("/updates", 200)
-	assert.Equal(t, `{"tag1":["update1","update2"],"tag2":["update1"]}`, s(data))
+	assert.Equal(t, map[string][]string{"tag1": {"update1", "update2"}, "tag2": {"update1"}}, updateNames(data))
 
 	data = tc.GET("/updates/tag1", 200)
-	assert.Equal(t, `{"tag1":["update1","update2"]}`, s(data))
+	assert.Equal(t, map[string][]string{"tag1": {"update1", "update2"}}, updateNames(data))
 	data = tc.GET("/updates/tag2", 200)
-	assert.Equal(t, `{"tag2":["update1"]}`, s(data))
+	assert.Equal(t, map[string][]string{"tag2": {"update1"}}, updateNames(data))
 	data = tc.GET("/updates/tag4", 200) // tag not exists
-	assert.Equal(t, "{}", s(data))
+	assert.Equal(t, map[string][]string{}, updateNames(data))
 
 	// Synthetic tag validation - create a bad tag on disk - request must still return 404
 	require.Nil(t, tc.fs.Updates.Rollouts.WriteFile("bad^tag", "update42", "rollout1", "foo"))
--- a/storage/api/api_storage.go
+++ b/storage/api/api_storage.go
@@ -126,6 +126,8 @@ type Storage struct {
 	stmtDeviceList      map[OrderBy]stmtDeviceList
 	stmtDeviceSetLabels stmtDeviceSetLabels
 	stmtDeviceSetUpdate stmtDeviceSetUpdate
+	stmtUpdateInsert    stmtUpdateInsert
+	stmtUpdateList      stmtUpdateList
 }
 
 func (d Device) Delete() error {
@@ -199,6 +201,8 @@ func NewStorage(db *storage.DbHandle, fs *storage.FsHandle) (*Storage, error) {
 		&handle.stmtDeviceGetLabels,
 		&handle.stmtDeviceSetLabels,
 		&handle.stmtDeviceSetUpdate,
+		&handle.stmtUpdateInsert,
+		&handle.stmtUpdateList,
 	); err != nil {
 		return nil, err
 	}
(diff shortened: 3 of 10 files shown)


storage/file_updates.go
```diff
@@ -219,13 +241,13 @@ func (s RolloutsFsHandle) ListUpdates(tag string) (map[string][]string, error) {
...
-	res := make(map[string][]string, len(tagDirs))
+	res := make(map[string][]Update, len(tagDirs))
 	for _, tag = range tagDirs {
 		if dirs, err := os.ReadDir(filepath.Join(s.root, tag)); err == nil {
-			res[tag] = make([]string, 0, len(dirs))
+			res[tag] = make([]Update, 0, len(dirs))
 			for _, d := range dirs {
 				if d.IsDir() {
-					res[tag] = append(res[tag], d.Name())
+					res[tag] = append(res[tag], s.readUpdateMeta(tag, d.Name()))
```
vkhoroz: There can be many updates for a single file, and this single line introduces an `O(n)` I/O complexity into a list function; which is barely scalable.

A more scalable approach is to have a single journal file for all tag's updates metadata, similar to what we have for configs etc, which has an O(1) complexity.
A trade-off is a bit more code to read metadata for a single update. But, it is incomparably more expensive to read multiple small files than to find a line inside a single file of a combined size.

storage/file_updates.go
```diff
@@ -235,6 +257,23 @@ func (s RolloutsFsHandle) ListUpdates(tag string) (map[string][]string, error) {
...
+		var meta updateMeta
+		if json.Unmarshal(data, &meta) == nil {
+			u.UploadedAt = meta.UploadedAt
+			return u
+		}
+	}
+	// Fall back to directory mtime for updates that predate metadata.json.
+	if info, err := os.Stat(filepath.Join(s.root, tag, name)); err == nil {
+		u.UploadedAt = info.ModTime().Unix()
+	}
```
vkhoroz: There are enough breaking changes made recently. So I think we can skip backward compatibility here.

cli/subcommands/updates/list.go
```diff
@@ -27,11 +29,11 @@ func listUpdates(api *api.Api) {
 	allUpdates, err := api.Updates().List()
 	cobra.CheckErr(err)
 
-	t := subcommands.NewTableWriter([]string{"TAG", "NAME"})
+	t := subcommands.NewTableWriter([]string{"TAG", "NAME", "UPLOADED AT"})
 
 	for tag, updates := range allUpdates {
 		for _, update := range updates {
-			t.AddRow(tag, update.Name)
+			t.AddRow(tag, update.Name, time.Unix(update.UploadedAt, 0).Format("2006-01-02 15:04:05"))
```
vkhoroz: Why not use `time.RFC3339` format?

storage/file_updates.go
```diff
@@ -219,22 +244,43 @@ func (s RolloutsFsHandle) ListUpdates(tag string) (map[string][]string, error) {
...
+			res[tag] = make([]Update, 0, len(dirs))
 			for _, d := range dirs {
 				if d.IsDir() {
-					res[tag] = append(res[tag], d.Name())
+					res[tag] = append(res[tag], s.readUpdateMeta(tag, d.Name()))
 				}
 			}
+			slices.SortFunc(res[tag], func(a, b Update) int {
+				return cmp.Compare(a.UploadedAt, b.UploadedAt)
+			})
```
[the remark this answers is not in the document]
vkhoroz: Use https://pkg.go.dev/slices#SortedStableFunc, which will do exactly the same thing Copilot suggests, as dir names are sorted by name.

If you instead agree to use the journal file, there is no sorting needed, as the append-only journal file lines are already sorted by date.

storage/file_updates.go
doanac: I did some performance testing. With a completely cold cache, listing 5000 updates on a single tag takes about 336ms. With a warm cache, it can go down to as low as 27ms.

That seems acceptable given 5000 is a pretty extreme value. However, we also have v2 plans to use storage buckets for things like this. In that model this would fall apart.

I'm going to rework this PR to use a sqlite table. That is better than the journal because it can also handle deletions which we'll be getting to eventually when we add "prune" support for getting rid of old updates

storage/file_updates.go
doanac: this was removed during the re-work of this feature

cli/subcommands/updates/list.go
doanac: fixed in force-push

storage/file_updates.go
doanac: I've just pushed a completely new version that uses the DB for this. I'm still testing, but I don't expect any substantial changes.

storage/api/api_storage.go
```diff
@@ -697,3 +704,49 @@ func (s *stmtDeviceDelete) run(uuid string) error {
...
+	return
+}
+
+func (s *stmtUpdateList) run(tag string) (map[string][]string, error) {
+	rows, err := s.Stmt.Query(tag, tag)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close() //nolint:errcheck
+	res := map[string][]string{}
```
vkhoroz: Could you use a `make(map...)` just for readability?

storage/file_updates.go
```diff
@@ -201,40 +201,6 @@ type RolloutsFsHandle struct {
 	UpdatesFsHandle
 }
 
-func (s RolloutsFsHandle) ListUpdates(tag string) (map[string][]string, error) {
```
vkhoroz, line 204: You seem to remove this function here, and then copy its code to backfill the updates from disk.

I find it reasonable to keep this function and use it for the backfill.
It is also useful for updates:
- After you call an API to insert updates, it makes sense to check both the database and the disk.

storage/db.go
```diff
@@ -157,6 +157,14 @@ func createTables(db *sql.DB) error {
 			scopes         TEXT,
 			FOREIGN KEY(user_id) REFERENCES user(id)
 		) WITHOUT ROWID;
+
+		CREATE TABLE updates (
+			tag         VARCHAR(80) NOT NULL,
+			name        VARCHAR(80) NOT NULL,
+			uploaded_at INT NOT NULL DEFAULT 0,
+			uploaded_by TEXT NOT NULL DEFAULT "",
```
vkhoroz, line 165: Maybe, `VARCHAR(200)` or something else limited?

storage/file_tar.go
```diff
@@ -217,6 +220,14 @@ func (s tarFsHandle) _handleTmpDir(destDirPath string, cfg tarUnpackConfig, unpa
 		}
 		return err
 	}
+	if cfg.events.onCommit != nil {
```
vkhoroz: I don't quite get what is being fixed by this.

IIRC, updates tarball upload should fail on same name error.
Thus, if it succeeds - DB insert must succeed too; unless the database was corrupted.
If, instead, we consider the filesystem to be our source of truth - I'd use upsert instead of an insert on the DB level.

FWIW there is no atomicity added here: the process can still crash, and the DB vs FS can still go out of sync.
I'd rather invest into assuming a fact that they can be out of sync.
IIUC the implementation without this commit does just that: it assumes that there is no update unless it is present in the database.

storage/api/api_storage.go
```diff
@@ -217,9 +221,65 @@ func NewStorage(db *storage.DbHandle, fs *storage.FsHandle) (*Storage, error) {
...
 	}
 
+	if err := handle.backfillUpdates(); err != nil {
+		return nil, err
+	}
+
 	return &handle, nil
 }
 
+func (s Storage) backfillUpdates() error {
```
vkhoroz: IIUC this is an attempt to fill the non-atomic way of the updates upload:
- If the process crashes, we might have an uploads file without the database entry, right?

Question: is there any other reason when they can get out of sync?
If yes - maybe, we should make this backfill logic a daemon thread.

At startup, I'd still run this in a separate goroutine to unblock the process initialization.

storage/api/api_storage.go
```diff
@@ -217,9 +221,65 @@ func NewStorage(db *storage.DbHandle, fs *storage.FsHandle) (*Storage, error) {
...
+	if err != nil {
+		return fmt.Errorf("backfill: failed to list DB updates: %w", err)
+	}
+	inDB := make(map[string]struct{})
+	for tag, updates := range existing {
+		for _, u := range updates {
+			inDB[tag+"/"+u.Name] = struct{}{}
+		}
+	}
+
```
vkhoroz: I'd use a key struct (a safer option).
Also, a bool simply feels more readable than a `struct{}` and uses as much or less space.
```suggestion
	type key struct { tag string; name string }
	inDB := make(map[key]bool)
	for tag, updates := range existing {
		for _, u := range updates {
			inDB[key{tag, u.Name}] = true
		}
	}
```

storage/api/api_storage.go
```diff
@@ -752,3 +812,16 @@ func (s *stmtUpdateList) run(tag string) (map[string][]Update, error) {
 	}
 	return res, rows.Err()
 }
+
+type stmtUpdateBackfill storage.DbStmt
+
+func (s *stmtUpdateBackfill) Init(db storage.DbHandle) (err error) {
+	s.Stmt, err = db.Prepare("apiUpdateBackfill", `
+		INSERT OR IGNORE INTO updates(tag, name, uploaded_at, uploaded_by) VALUES(?, ?, ?, ?)`)
```
vkhoroz: Why can't we reuse the `InsertUpdate`?
IIUC the backfill daemon only tries to insert the updates which are missing from the database; so there must be no collisions.

vkhoroz (commented):
Seems good.

I only had a few small comments.

storage/file_updates.go
vkhoroz: When I took a deeper look at your backfill code, it seems different.
So, feel free to ignore this comment if you wish.

storage/db.go
doanac: we have the username set to the same type higher up the file

storage/file_tar.go
doanac: This is to handle the case that the write to the DB fails. If we can't commit to DB, we should try and remove this. It's not perfect but does help

storage/api/api_storage.go
doanac: Its more about hitting something where you try and restore from a backup. Claude liked the idea and it was handy for my scenario. However, if its going to cause confusion - i'd vote to drop this commit from the PR

storage/api/api_storage.go
doanac: The uploaded_at timestamp would be wrong. We can change and have the primary API require a timestamp and then have one function, but I think the direction this is headed, we should just drop this backfill stuff

storage/api/api_storage.go
doanac: Can you explain how its safer?

As per bool vs struct{}  - I was a "bool" person forever, but it seems this isn't the way people do it and there are claims its more obvious - `struct{}` means you are dealing with a set. `bool` means a dict of bools.

storage/api/api_storage.go
vkhoroz: I'm fine to drop this commit, and also the "onCommit" thing.
Both seem to be an overengineering at this time to me:
- The "onCommit" assumes that the database commit can fail _after_ a successful file write. But I see no use case when that is possible, and you seem to provide no evidence where it helps in practice. Still, it does not solve the hard crash, meaning that it doesn't even help with the problem it claims to solve.
- The "backfill" tries to solve the same exact problem in a different way i.e. it is a competing, not substituting commit. Although it is more appropriate, as it covers a hard crash, I lean towards dropping it to due to its complexity. The restore from backup seems interesting, but why would anyone backup file system without the database?

If you drop these two commits, I have zero comments.

storage/file_tar.go
vkhoroz: See below.

storage/api/api_storage.go
vkhoroz: It's safer in a way that it does not assume about the format of tag and username, in particular that "/" is not a part of either constituent.

storage/api/api_storage.go
doanac: > The "onCommit" assumes that the database commit can fail after a successful file write. But I see no use case when that is possible

The call to commit the change to the DB can return an error.  In the event it returns an error - what should we do?

 * If we ignore the error - the update won't exist when we try to list it. But the user won't be able to create the update because it exists on disk.
 * If we return the error - the update still exists on disk, so the user can't retry the operation.

doanac: what i'll do is this:
 * pre-commit the change to the db
 * the List api requires both the DB entry *and* the directory to exist.

doanac: I've force pushed a new version that may resolve our problems

storage/api/api_storage.go
```diff
@@ -319,8 +325,73 @@ func (s Storage) ReadAppliedConfigs(uuid string) (*storage.AppliedConfigs, error
...
+					dbUpdates[tag] = slices.Delete(tags, idx, idx+1)
+				}
+			}
+		}
+	}
+
+	return dbUpdates, nil
+}
+
+func (s Storage) listUpdatesFromFS(tag string) (updateSet, error) {
```
vkhoroz: I think you could reuse the previous `func (s RolloutsFsHandle) ListUpdates(tag string) (map[string][]string, error)`.

IMHO, there is no need for that fancy `updateKey` struct, as your `dbUpdates` loop could be written in a more optimal way using a regular map of arrays (see my other comment).

storage/api/api_storage.go
```diff
@@ -319,8 +325,73 @@ func (s Storage) ReadAppliedConfigs(uuid string) (*storage.AppliedConfigs, error
...
+				tags := dbUpdates[tag]
+				if len(tags) == 1 {
+					delete(dbUpdates, tag)
+				} else {
+					idx := slices.Index(tags, u)
+					dbUpdates[tag] = slices.Delete(tags, idx, idx+1)
+				}
+			}
+		}
+	}
```
vkhoroz: You might have used a simpler inFS structure (`map[string][]string`) to early exit for non-existent tag.
Also, deleting non-existent updates can be implemented more efficiently.

```suggestion
	// [Change] Use a function you deleted in this commit.
	fsUpdates, err := s.fs.Updates.Rollouts.ListUpdates(tag)
	// ...
	for tag, updates := range dbUpdates {
		// [Change] Search for a tag in FS updates list to delete early and skip the inner loop.
		fsTagUpdates, ok := fsUpdates[tag]
		if !ok {
			for _, u := range updates {
				slog.Warn("Found update in DB but not in FS, removing from list", "tag", tag, "update", u)
			}
			delete(dbUpdates, tag)
			continue
		}
		// [Change] Iterate in reverse order for more efficient deletions (no search needed).
		for idx := len(updates) - 1; idx >= 0; idx-- {
			u := updates[idx]
			// [Change] Leverage the fs updates list for this specific tag.
			// As FS returns sorted dir/file names, we can use a binary search (complexity comparable to a map).
			if _, ok = slices.BinarySearch(fsTagUpdates, u); !ok {
				slog.Warn("Found update in DB but not in FS, removing from list", "tag", tag, "update", u)
				updates = slices.Delete(updates, idx, idx+1)
			}
		}
		// [Change] Amend dbUpdates outside the inner loop for efficiency.
		if len(updates) > 0 {
			dbUpdates[tag] = updates
		} else {
			delete(dbUpdates, tag)
		}
	}
```

vkhoroz (commented):
This is simple enough now.

storage/api/api_storage.go
vkhoroz: I still don't get why would we ever receive a database error for this operation, other than a complete disaster.

Anyway, your new commit handles all error cases in a nice way, so it makes sense.

storage/api/api_storage.go
doanac: I don't want to keep a public function around that could get used by accident. As for efficiency - we are talking about a small chance of this happening for a small data set. Let's keep it short and simple rather than optimized.

storage/api/api_storage.go
vkhoroz: There is still a problem that a user would not be able to re-upload a failed update - something that was provided by the previous solution (before adding the database layer)

I actually have an idea how to solve that problem in a different way.
Please, walk the dog on it.

**TL;DR**
1. Before the upload look into the database to see if the update exists.  If yes - fail with conflict.
2. If the update does not exists in the database - we can safely upload and replace whatever was on the disk.
    There is an option in `file_tar.go` for that; and the whole upload operation is actually atomic (directory rename).
3. Insert the record into the database _after_ a successful upload.

Now, if the upload fails, a user can simply re-upload, and it will succeed, as there is no record in the database.

There can still be a garbage on disk, and that's a trade-off, but the probability is extremely low, and we can never get it to zero due to the application crash possibility.

doanac: A user can re-upload now. The code ignores a failed db insert if the entry already existed. Then it goes to the file upload logic which was already doing the correct things.

vkhoroz: hmm... It means we're fine.
We could have a faster listing of updates and less code if we used [this option](https://github.com/foundriesio/update-server/blob/main/storage/file_tar.go#L71).
Anyway, feel free to go as you wish. Your current PR does it good enough in a more complex way.

vkhoroz: I mean: it's usually nice to use use an option which is already implemented and allows to get rid of ~80 lines of new code.

doanac: Let's go with your idea. Please force-push your approach onto this branch. I'll merge it Monday morning.

vkhoroz: force-pushed a simpler way to implement that two-phased commit idea.

storage/api/api_storage.go
vkhoroz: With a different two-phase commit we no longer need to list updates from the filesystem.

storage/file_updates.go
```diff
@@ -87,7 +93,7 @@ func (s updatesFsHandleWrap) SaveUpload(tag, update string, payload io.Reader, o
 	destDir = filepath.Join(destDir, tag, update)
 	h := tarFsHandle{root: root}
 	return h.unpackTar(payload, destDir,
-		TarUnpackReplaceDest(false), // Fail if the update with the same tag and name already exists.
+		TarUnpackReplaceDest(true), // Replace updates with the same tag and name - uniqueness is checked on the database level.
 		TarUnpackUseTmpFile("update.tar"),
 		TarUnpackUseTmpDir(txDir),
```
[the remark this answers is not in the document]
vkhoroz: This is by design:
- Filesystem corruption is impossible due to an upload/unpack into a temporary directory and an atomic directory move.
- Overwrite is there to allow users re-upload a failed upload.
- Two-phase commit db->fs->db prevents conflicts unless there are concurrent uploads for the same tag and name.  This is an acceptable trade-off.

storage/api/api_storage.go
```diff
@@ -500,12 +506,25 @@ func (s Storage) UploadConfigs(payload io.Reader) (err error) {
...
 	cleanup := func(cleanupErr error) {
 		// This is not critical - log and let the "real" error/success return below.
 		slog.Error("Failed to clean upload directory", "error", cleanupErr)
 	}
-	return s.fs.Updates.SaveUpload(tag, updateName, payload, cleanup)
+	if err := s.fs.Updates.SaveUpload(tag, updateName, payload, cleanup); err != nil {
+		return err
+	}
+	return s.stmtUpdateInsert.run(tag, updateName, uploadedBy)
 }
```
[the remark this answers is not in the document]
vkhoroz: Unique constraint is checked by a database query before the filesystem write. This is a way we implemented a two-phase commit - by design.
Another option was to run fs change inside the database transaction - but that would block the entire database for writes for the entire upload duration - the worst possible solution.

storage/db.go
```diff
@@ -157,6 +157,14 @@ func createTables(db *sql.DB) error {
...
+			PRIMARY KEY (tag, name)
+		) WITHOUT ROWID;
```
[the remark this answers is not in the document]
vkhoroz: this is by design atm

server/ui/api/handlers_test.go
```diff
@@ -604,23 +604,43 @@ func TestApiUpdateList(t *testing.T) {
...
 	require.Nil(t, tc.fs.Updates.Rollouts.WriteFile("tag2", "update1", "rollout1", "foo"))
 	require.Nil(t, tc.fs.Updates.Rollouts.WriteFile("tag2", "update3", "rollout1", "foo"))
 
 	data := tc.GET("/updates", 200)
-	assert.Equal(t, `{"tag1":["update1","update2"],"tag2":["update1","update3"]}`, s(data))
+	assert.Equal(t, map[string][]string{"tag1": {"update1", "update2"}, "tag2": {"update1"}}, updateNames(data))
+
+	// now add rollout for update3 - it should now appear in the list of updates for tag2
+	require.Nil(t, tc.fs.Updates.Rollouts.WriteFile("tag2", "update3", "rollout1", "foo"))
+
```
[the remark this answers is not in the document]
vkhoroz: fixed

storage/api/api_storage.go
```diff
@@ -319,8 +324,9 @@ func (s Storage) ReadAppliedConfigs(uuid string) (*storage.AppliedConfigs, error
 
 var clearingEventTypes = []string{"EcuInstallationCompleted", "CertRotationCompleted", "MetadataUpdateCompleted"}
 
-func (s Storage) ListUpdates(tag string) (map[string][]string, error) {
-	return s.fs.Updates.Rollouts.ListUpdates(tag)
+// ListUpdates returns a map of tag to update names. If the tag is empty, it returns all tags.
+func (s Storage) ListUpdates(tag string) (map[string][]Update, error) {
+	return s.stmtUpdateList.run(tag)
 }
```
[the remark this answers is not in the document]
vkhoroz: fixed

PR #185, opened 2026-06-18
feat: manage deleted devices (list + restore)

StealthyCoder:
## What

Implements #182 — a way to **list** and **remove** devices from the "deleted" list, across the storage, REST, and CLI layers, plus docs.

Deleting a device (`DELETE /devices/{uuid}`) soft-deletes it (`deleted=1`) and stops it being served. Until now there was no way to see which devices are in that state or to bring one back. This adds:

- **REST**
  - `GET /v1/devices/deleted` — list soft-deleted devices; paginated/sortable exactly like `GET /devices`. Scope: `devices:read`.
  - `POST /v1/devices/{uuid}/restore` — remove a device from the deleted list (clears the tombstone). Idempotent. Scope: `devices:delete`.
- **CLI**
  - `fiocli devices list-deleted` — mirrors `devices list` (same columns/sort/paging).
  - `fiocli devices restore <uuid>`.
- **Docs** — `docs/api.md` updated (Swagger is generated from the handler annotations at release time).

## Design notes

- **Separate `GET /devices/deleted` collection rather than a `?deleted=true` param.** Every device statement (`stmtDeviceGet`/`List`/`Count`) hard-filters `WHERE deleted=false`, and the list statements are pre-built per sort order in a `map[OrderBy]stmtDeviceList`. A query param would force those clean statements to become conditional/dynamic; a parallel statement set is purely additive. Echo's router matches the static `deleted` segment before `:uuid`, so there's no route shadowing.
- **`POST .../restore`, idempotent.** `UPDATE devices SET deleted=0 WHERE uuid=?` with no `AND deleted=1` guard. `RowsAffected > 0` → `204` (uuid exists, whether it was deleted or already active); `0` → `404` (unknown uuid). A retry after a dropped `204` returns `204`, not a spurious `404`. (Relies on SQLite's matched-rows semantics for `RowsAffected` — noted at the call site.)
- **Scopes:** list = `devices:read` (like `GET /devices`); restore = `devices:delete` (the inverse of delete). No new scope introduced.

## ⚠️ Restore is DB-record-only

`Device.Delete()` sets `deleted=1` **and** destroys the device's filesystem data (`os.RemoveAll` of configs, update events, apps-states). Restore therefore recovers the **server record only** — that history is not recovered. This is documented in a `Device.Delete()` doc comment and in `docs/api.md`; `Delete()` behaviour is intentionally **not** changed here.

**Order-of-operations gotcha (documented):** the restored record keeps the device's pre-deletion public key, and the gateway does not support key rotation. A device that regenerated its key *while deleted* (e.g. after a factory reset) cannot be restored — once restored it presents a new key that no longer matches the stored one, and because a record still exists the gateway won't auto-enrol it either. Such a device must be deleted and re-enrolled from scratch.

## Boy-scout fixes

- Added `DEFAULT 0` to the `devices.deleted` column (it had no default, unlike `users.deleted`).
- Documented `Device.Delete()` as destructive.

## Suggested follow-up

Make `Device.Delete()` non-destructive — retain filesystem data behind the `deleted=1` tombstone instead of `os.RemoveAll` — so restore becomes lossless. Supporting gateway key rotation (the commented-out `RotatePubKey` path) would additionally make restore robust to key changes.

## Commits

Six small, independently-reviewable commits (each builds + tests green, for clean `git bisect`):

1. `fix:` add `DEFAULT 0` to `devices.deleted` column
2. `refactor:` document `Device.Delete()` as destructive
3. `feat:` storage support for listing and restoring deleted devices
4. `feat:` REST endpoints to list and restore deleted devices
5. `feat:` CLI commands to list and restore deleted devices
6. `docs:` document deleted-device list and restore endpoints

## Testing

- `go build ./...` and `make fioserver` — clean.
- `go test ./...` — all packages pass, including new `TestDeviceRestore`, `TestApiDeviceRestore`, and `TestApiDeviceListDeleted`.
- Swagger annotations validated with `make swagger`; both new paths appear in the generated spec.

Closes #182

### DIFF
--- a/cli/api/devices.go
+++ b/cli/api/devices.go
@@ -90,6 +90,17 @@ func (d DeviceApi) Delete(uuid string) error {
 	return d.api.Delete(fmt.Sprintf("/v1/devices/%s", uuid))
 }
 
+// ListDenied returns all UUIDs on the denied list.
+func (d DeviceApi) ListDenied() ([]string, error) {
+	var uuids []string
+	return uuids, d.api.Get("/v1/denied-devices", &uuids)
+}
+
+// DeleteDenied removes a device from the denied list so it can access the backend again.
+func (d DeviceApi) DeleteDenied(uuid string) error {
+	return d.api.Delete(fmt.Sprintf("/v1/denied-devices/%s", uuid))
+}
+
 func (d *DeviceApi) Tests(uuid string) ([]TargetTest, error) {
 	var tests []TargetTest
 	return tests, d.api.Get(fmt.Sprintf("/v1/devices/%s/tests", uuid), &tests)
--- a/server/ui/api/handlers.go
+++ b/server/ui/api/handlers.go
@@ -39,6 +39,8 @@ func RegisterHandlers(e *echo.Echo, storage *storage.Storage, a auth.Provider) {
 	g.GET("/devices", h.deviceList, requireScope(users.ScopeDevicesR))
 	g.GET("/devices/:uuid", h.deviceGet, requireScope(users.ScopeDevicesR))
 	g.DELETE("/devices/:uuid", h.deviceDelete, requireScope(users.ScopeDevicesD))
+	g.GET("/denied-devices", h.deniedDevicesList, requireScope(users.ScopeDevicesR))
+	g.DELETE("/denied-devices/:uuid", h.undenyDevice, requireScope(users.ScopeDevicesD))
 	g.GET("/devices/:uuid/apps-states", h.deviceAppsStatesGet, requireScope(users.ScopeDevicesR))
 	g.GET("/devices/:uuid/tests", h.deviceTestsList, requireScope(users.ScopeDevicesR))
 	g.GET("/devices/:uuid/tests/:testid", h.deviceTestGet, requireScope(users.ScopeDevicesR))
--- a/storage/api/api_storage.go
+++ b/storage/api/api_storage.go
@@ -131,6 +131,10 @@ type Storage struct {
 	stmtUpdateList      stmtUpdateList
 }
 
+// Delete is DESTRUCTIVE. It both adds the device to the denied list in the DB
+// (deleted=1) and removes the device's filesystem data (configs, update
+// events, apps-states). The filesystem data destroyed here is NOT recovered
+// if the device is later allowed back via UndenyDevice.
 func (d Device) Delete() error {
 	err1 := d.storage.stmtDeviceDelete.run(d.Uuid)
 	err2 := d.storage.fs.Devices.Delete(d.Uuid)
--- a/storage/api/api_storage.go
+++ b/storage/api/api_storage.go
@@ -119,16 +119,18 @@ type Storage struct {
 	db *storage.DbHandle
 	fs *storage.FsHandle
 
-	stmtDeviceCount     stmtDeviceCount
-	stmtDeviceDelete    stmtDeviceDelete
-	stmtDeviceGet       stmtDeviceGet
-	stmtDeviceGetGroups stmtDeviceGetGroups
-	stmtDeviceGetLabels stmtDeviceGetLabels
-	stmtDeviceList      map[OrderBy]stmtDeviceList
-	stmtDeviceSetLabels stmtDeviceSetLabels
-	stmtDeviceSetUpdate stmtDeviceSetUpdate
-	stmtUpdateInsert    stmtUpdateInsert
-	stmtUpdateList      stmtUpdateList
+	stmtDeviceCount      stmtDeviceCount
+	stmtDeviceDelete     stmtDeviceDelete
+	stmtDeviceGet        stmtDeviceGet
+	stmtDeviceGetGroups  stmtDeviceGetGroups
+	stmtDeviceGetLabels  stmtDeviceGetLabels
+	stmtDeviceList       map[OrderBy]stmtDeviceList
+	stmtDeniedDeviceList stmtDeniedDeviceList
+	stmtUndenyDevice     stmtUndenyDevice
+	stmtDeviceSetLabels  stmtDeviceSetLabels
+	stmtDeviceSetUpdate  stmtDeviceSetUpdate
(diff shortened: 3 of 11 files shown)


storage/api/api_storage.go
```diff
@@ -128,6 +128,12 @@ type Storage struct {
 	stmtDeviceSetUpdate stmtDeviceSetUpdate
 }
 
+// Delete is DESTRUCTIVE. It both marks the device as deleted in the DB
+// (deleted=1) and removes the device's filesystem data (configs, update
+// events, apps-states). A device can later be restored with DeviceRestore,
+// but that only clears the DB tombstone - the filesystem data destroyed here
+// is NOT recovered. TODO: make deletion non-destructive so restore is lossless
```
doanac: Remove the TODO - Claude is thinking this feature is like a soft delete. It's not really that.

storage/api/api_storage.go
```diff
@@ -244,6 +256,41 @@ func (s Storage) DevicesList(opts DeviceListOpts) ([]DeviceListItem, int, error)
 	return devices, total, nil
 }
 
+// DevicesListDeleted lists devices that have been soft-deleted (the "deleted"
+// list). It mirrors DevicesList but reads from the deleted statement set
+// (WHERE deleted=true).
+func (s Storage) DevicesListDeleted(opts DeviceListOpts) ([]DeviceListItem, int, error) {
```
doanac: This is starting to deviate too far from our current SAAS backend. This PR is treating the feature like a "soft-delete" feature. This feature about managing "denied devices". Due to mTLS - we could:

 * Delete a device from the DB
 * The device makes an mTLS connection
 * The backend will see this as a new device, create an entry and give it access to the backend.

We keep a delete=true flag to prevent this scenario. So we wind up with a list of "denied devices" as our SAAS calls them. Every now and then an operator makes a mistake and needs that device (with the exact same cert and UUID) to connect back to the product. This is where we need to remove a device from the denied list.

doanac: This is a long winded way of saying - this function doesn't need sorting or anything like that. It just needs to return one giant array of UUIDs.

server/ui/api/handlers.go
```diff
@@ -37,8 +37,10 @@ func RegisterHandlers(e *echo.Echo, storage *storage.Storage, a auth.Provider) {
 	g.GET("/configs/device/:uuid/applied", h.deviceAppliedConfigsGet, requireScope(users.ScopeDevicesR))
 	g.GET("/configs/device/:uuid/history", h.deviceConfigsHistory, requireScope(users.ScopeDevicesR))
 	g.GET("/devices", h.deviceList, requireScope(users.ScopeDevicesR))
+	g.GET("/devices/deleted", h.deviceListDeleted, requireScope(users.ScopeDevicesR))
 	g.GET("/devices/:uuid", h.deviceGet, requireScope(users.ScopeDevicesR))
 	g.DELETE("/devices/:uuid", h.deviceDelete, requireScope(users.ScopeDevicesD))
+	g.POST("/devices/:uuid/restore", h.deviceRestore, requireScope(users.ScopeDevicesD))
```
doanac: The APIs here should be more like our SAAS apis - we are managing a "denied" list so:
```
   GET /denied-devices
   DELETE /denied-devices/:uuid
```

cli/api/devices.go
```diff
@@ -31,8 +31,20 @@ func (a *Api) Devices() DeviceApi {
 // ListPage fetches a single page of devices. It returns the devices,
 // whether more pages are available, and the total number of pages.
 func (d DeviceApi) ListPage(page int, limit int, sortBy string) ([]DeviceListItem, bool, int, error) {
```
doanac, line 33: The CLI should mimic fioctl and do:
```
 fiocli devices list-denied
 fiocli devices delete-denied
```

doanac (commented):
This needs some rework. Its implementing and trying to clarify nuances of "soft-delete" functionality. This isn't a soft-delete feature. This is about managing a list of denied devices. When a device is deleted - it needs to be moved into a list of denied devices (we choose deleted=True in the DB so we have its UUID and pubkey) to prevent that devices from accessing the backend with mTLS.

cli/api/devices.go
```diff
@@ -31,8 +31,14 @@ func (a *Api) Devices() DeviceApi {
...
+	return d.listPage("/v1/devices", page, limit, sortBy)
+}
+
+// listPage fetches a single page from a device-listing resource, returning the
+// devices, whether more pages are available, and the total number of pages.
+func (d DeviceApi) listPage(resource string, page int, limit int, sortBy string) ([]DeviceListItem, bool, int, error) {
 	offset := (page - 1) * limit
-	resource := fmt.Sprintf("/v1/devices?limit=%d&offset=%d", limit, offset)
+	resource = fmt.Sprintf("%s?limit=%d&offset=%d", resource, limit, offset)
 	if sortBy != "" {
```
doanac, line 36: I think this may be left over from a version where you were going to have pagination. I'm remove this.

cli/subcommands/devices/list.go
```diff
@@ -87,7 +87,13 @@ func validateColumns(columnsStr string) ([]string, error) {
...
 	devices, hasMore, totalPages, err := dapi.ListPage(page, defaultPageLimit, sortBy)
 	cobra.CheckErr(err)
+	renderDeviceList(devices, columns, page, hasMore, totalPages)
+}
 
+// renderDeviceList prints a page of devices as a table, followed by a
+// pagination hint when more pages are available. It is shared by the `list`
+// and `list-deleted` commands.
+func renderDeviceList(devices []api.DeviceListItem, columns []string, page int, hasMore bool, totalPages int) {
 	headers := make([]string, 0, len(columns))
```
doanac, line 91: This may be left from an attempt at pagination that you abandoned.

doanac (commented):
This is looking much better. I only have a couple small complaints

doanac (approved):

PR #186, opened 2026-06-25
Add support for oauth2 device flow

doanac:
This set of changes adds support for oauth2 device flow to streamline the CLI login process. It also sets the basis for followup work to support lmp-device-register

### DIFF
--- a/auth/csrf.go
+++ b/auth/csrf.go
@@ -46,6 +46,10 @@ func CsrfCheck(next echo.HandlerFunc) echo.HandlerFunc {
 			return next(c)
 		}
 
+		if c.Path() == "/oauth2/authorization/device/" || c.Path() == "/oauth2/token/" {
+			return next(c)
+		}
+
 		cookie, err := c.Cookie(CsrfCookieName)
 		if err != nil || cookie.Value == "" {
 			return c.String(http.StatusForbidden, "Missing CSRF cookie")
--- a/server/ui/api/handlers_oauth2.go
+++ b/server/ui/api/handlers_oauth2.go
@@ -1,0 +1,210 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+package api
+
+import (
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/labstack/echo/v4"
+
+	"github.com/foundriesio/update-server/storage/users"
+)
+
+type oauth2Handlers struct {
+	users *users.Storage
+}
+
+type DeviceCodeRequest struct {
+	Scopes       string `form:"scope"`
+	TokenExpires int64  `form:"token_expires"`
+}
+
+type DeviceCodeResponse struct {
+	DeviceCode              string `json:"device_code"`
+	UserCode                string `json:"user_code"`
+	VerificationURI         string `json:"verification_uri"`
+	VerificationURIComplete string `json:"verification_uri_complete"`
+	ExpiresIn               int    `json:"expires_in"` // seconds until expiry, per RFC 8628
+	Interval                int    `json:"interval"`
+}
+
+type DeviceTokenRequest struct {
+	DeviceCode string `form:"device_code"`
+	GrantType  string `form:"grant_type"`
+}
+
+type DeviceTokenResponse struct {
+	AccessToken string `json:"access_token"`
+	TokenType   string `json:"token_type"`
+	Expires     int64  `json:"expires"`
+	Scopes      string `json:"scope"`
+}
+
+type oauth2Error struct {
+	Error            string `json:"error"`
+	ErrorDescription string `json:"error_description,omitempty"`
+}
+
+// @Summary Initiate OAuth2 Device Authorization
+// @Accept json
+// @Param data body DeviceCodeRequest true "Device code request"
+// @Produce json
+// @Success 200
+// @Router  /oauth2/device/code [post]
+func (h oauth2Handlers) oauth2DeviceCode(c echo.Context) error {
+	var req DeviceCodeRequest
+	if err := c.Bind(&req); err != nil {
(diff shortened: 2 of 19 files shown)


server/ui/api/handlers_test.go
```diff
@@ -215,6 +215,14 @@ func (p testAuthProvider) GetSession(c echo.Context) (*auth.Session, error) {
 func (testAuthProvider) DropSession(echo.Context, *auth.Session) {
 }
 
+func (testAuthProvider) GetRateLimiterMiddleware() echo.MiddlewareFunc {
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) error {
+			return next(c)
+		}
```
vkhoroz: You could do `return next` (replace 3 lines with just 1) to achieve the same result.

storage/users/user_storage.go
```diff
@@ -48,6 +48,22 @@ func (u User) GetAuditLog() (string, error) {
 	return u.h.fs.Audit.ReadEvents(u.id)
 }
 
+func (u User) ApproveAuthorization(deviceCode string, tokenDescription string, scopes Scopes) error {
```
vkhoroz, line 51: These two functions can be a part of `oauth2.go` too.
I have no preference, just pointing at the possibility.

server/ui/api/handlers_oauth2.go
```diff
@@ -0,0 +1,215 @@
...
+	ErrorDescription string `json:"error_description,omitempty"`
+}
+
+// @Summary Initiate OAuth2 Device Authorization
+// @Accept json
+// @Param data body DeviceCodeRequest true "Device code request"
+// @Produce json
+// @Success 200
+// @Router  /oauth2/device/code [post]
+func (h oauth2Handlers) oauth2DeviceCode(c echo.Context) error {
```
vkhoroz, line 57: Why can't we reuse the `handlers` class for this?

I suspect you're trying to start spinning off individual handlers classes for different groups.
But, that would better be done as an overhaul; or we risk getting into a mess.

server/ui/api/handlers_oauth2.go
```diff
@@ -0,0 +1,215 @@
...
+	})
+}
+
+// @Summary Handle OAuth2 Device Token Polling
+// @Accept json
+// @Param data body DeviceTokenRequest true "Device token request"
+// @Produce json
+// @Success 200
+// @Router  /oauth2/device/token [post]
+func (h oauth2Handlers) oauth2DeviceToken(c echo.Context) error {
```
vkhoroz, line 115: Why didn't some standard server library (like e.g. https://pkg.go.dev/github.com/go-oauth2/oauth2/v4) suit our needs with some minimal customization?
For the above example, it looks like all we needed was to implement the store interface from https://github.com/go-oauth2/oauth2/blob/master/store.go and some minimal handlers customization.

Just asking, as I don't feel expert enough to validate the full correctness of this Oauth2 implementation.

auth/csrf.go
```diff
@@ -46,6 +46,10 @@ func CsrfCheck(next echo.HandlerFunc) echo.HandlerFunc {
 			return next(c)
 		}
 
+		if c.Path() == "/oauth2/device/code" || c.Path() == "/oauth2/device/token" {
+			return next(c)
```
vkhoroz, line 50: What's the problem with using CSRF for the Oauth2?

auth/provider_local.go
```diff
@@ -198,10 +199,12 @@ func (p localProvider) renderLoginPage(c echo.Context, reason string) error {
 		User      *users.User
 		NavItems  []string
 		CsrfToken string
+		Version   string
 	}{
 		Title:     "Login",
 		Reason:    reason,
 		CsrfToken: csrfToken,
+		Version:   version.Version,
```
vkhoroz: It looks like this is a standalone bug fix, right?
Maybe, a separate PR to commit faster?

auth/csrf.go
doanac: This is device/api-client side.

server/ui/api/handlers_oauth2.go
doanac: That library doesn't implement the device-flow part of the spec that we need. Chatting with claude/copilot, they both said this library doesn't buy us much. Especially since we have token generation, validation, and scope logic already implemented.

server/ui/api/handlers_oauth2.go
doanac: We need a slightly different middleware setup for this that:
 * doesn't require authenticated clients
 * does include some ratelimits (since it doesn't require authentication)

storage/users/user_storage.go
doanac: good idea. i applied this change as its how sessions.go was done

mike-scott:
NOTE: Do we need to worry about existing users getting an error once they move to this new version?

```
{"time":"2026-08-25T12:11:48.351875301-07:00","level":"ERROR","msg":"command failed","error":"failed to initialize users storage: unable to prepare 'oauth2DeviceAuthCreate' statement: no such table: oauth2_device_flow"}
```

mike-scott (approved):
I successfully tested this PR on top of `main` using real HW.

- Started with QLI 2.0 + Foundries components clean build (not registered)
- Use the following command to register the device against my local update-server
```
DEVICE_API=http://local-server:8080/v1/devices \
OAUTH_BASE=http://local-server:8080/oauth2 \
fio-device-register \
  --name mike-iq-9075-evk \
  --tags iq-9075-evk-main \
  --factory mike-iq-9075-evk
```
- After completing the OAUTH dance, the device checked into the server and was reporting
  `Local TUF repo is invalid: Could not load latest root` (no TUF root installed)
- Uploaded the same build's `ostree_repo` as an update and added it to a rollout
- The device checked in and correctly matched the update with it's currently installed rootfs.
- No more `Local TUF .. ` error message

NOTE: I have a contrib/e2e/ test which also exercises the OAUTH paths (fiocli and mock device register).  I can create a new PR once this is merged.

PR #188, opened 2026-06-26
Local setup with mock data

MiloCasagrande:
- **feat: local dev setup with mock data**
- **fix: add .local-data to .gitignore**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 docs/swagger/*
 bin/*
 release/*
+.local-data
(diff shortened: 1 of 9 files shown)


doanac (commented):
I like the general idea. We have a "contrib/gen-certs.sh" but I think its different enough that we shouldn't bother trying to get it aligned with this new seed command.

Two things:
1) @StealthyCoder - You are doing some hackery around seeding data for scale testing. Is there anything here that you could leverage or is what you are doing totally different?
2) Can you squash the commits where appropritate? I think commits 1,2 and 4 of this are probably a single change?

doanac (approved):

PR #193, opened 2026-07-01
UI branding

MiloCasagrande:
This is the work to brand the UI via a single `branding.json` file. There is a `branding.md` document to describe how to use the feature.

Dark mode works only if the browser/system is set into dark mode.

There are 4 commits, split into the different parts of the work (general branding, favicon, dark mode, branding doc), I can squash them and keep only 2 in case.

A couple of screenshots to visually see it:

ui-branded-light

ui-branded-dark

### DIFF
--- a/server/ui/web/branding.go
+++ b/server/ui/web/branding.go
@@ -1,0 +1,85 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+package web
+
+import (
+	"encoding/json"
+	"log/slog"
+	"regexp"
+)
+
+// cssColor allows hex, rgb()/rgba()/hsl() functional notation, and CSS named
+// colors — i.e. the characters those need. Anything else (e.g. ";", "{") is
+// rejected so a bad branding.json can't inject or break the stylesheet.
+var cssColor = regexp.MustCompile(`^[#a-zA-Z0-9(),.%\s/-]+$`)
+
+// Branding holds the resolved branding values (defaults applied) used by both
+// HTML templates and the templated CSS.
+type Branding struct {
+	Title      string // app name in topbar + <title> suffix
+	Logo       string // filename under <data-dir>/branding/, "" = text brand only
+	Primary    string // --brand-primary
+	Accent     string // --brand-accent
+	Surface    string // --surface-1 (page background)
+	SurfaceAlt string // --surface-2 (content background)
+	Text       string // --text-1
+}
+
+// brandingFile is the on-disk JSON shape. Pointers distinguish "absent" (keep
+// default) from "explicitly set".
+type brandingFile struct {
+	Title  *string `json:"title"`
+	Logo   *string `json:"logo"`
+	Colors struct {
+		Primary    *string `json:"primary"`
+		Accent     *string `json:"accent"`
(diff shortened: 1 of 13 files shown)


server/ui/web/handlers.go
```diff
@@ -20,27 +24,36 @@
...
+		provider:    authProvider,
+		templates:   templates.Templates,
+		branding:    branding,
+		brandingDir: brandingDir,
 	}
+	var rendered bytes.Buffer
+	if err := h.templates.ExecuteTemplate(&rendered, "style.css", branding); err != nil {
+		slog.Error("failed to render style.css for etag", "error", err)
+	}
+	h.styleEtag = fmt.Sprintf("%x", md5.Sum(rendered.Bytes()))
```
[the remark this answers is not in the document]
MiloCasagrande: Not sure here, I can change the hash algorithm to a more robust `sha256`: it's to calculate the hash of the CSS file though.

Not sure how fast/slow one would be compared to the other, maybe `sha256` is faster...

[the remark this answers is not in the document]
doanac: ignore this. md5 is fine for caching logic

server/ui/web/branding.go
```diff
@@ -0,0 +1,85 @@
...
+	Title  *string `json:"title"`
+	Logo   *string `json:"logo"`
+	Colors struct {
+		Primary    *string `json:"primary"`
+		Accent     *string `json:"accent"`
+		Surface    *string `json:"surface"`
+		SurfaceAlt *string `json:"surface-alt"`
+		Text       *string `json:"text"`
+	} `json:"colors"`
+}
```
doanac, line 61: I'm not sure I get the value of having two different structs? I think I'd stick with just the brandingFile idea but get rid of the pointers do length == 0 checks instead checks for nil.

doanac (commented):
This is cool. I have one nit - but maybe I'm wrong in my opinion.

server/ui/web/branding.go
MiloCasagrande: The pointers are overkill, yeah. I'll switch to plain strings and drop the nil checks.

MiloCasagrande: Regarding the two structs, you mean `brandingFile` and `Branding`?

In that case I think it's still worth keeping them both, as they serve different purposes.

`brandingFile` mirrors the raw JSON of the `branding.json` file used to customize the UI colors.

`Branding` is the flat version, template-friendly with defaults applied: this is guaranteed to be populated and valid via `LoadBranding`.

Staying only with brandingFile would lead to move the "default value resolution/validation" logic into the templates, and we might have to duplicate it in multiple places.

doanac: I was thinking more of having something like an "applyDefaults" method added to the brandingFile struct. Then pass that to the templates and everything would be in place for it?

If that approach looks bad - stick with this. I'm fine either way. Merge at your leisure.

MiloCasagrande: The only reasons I still prefer the two structs approach is that the only way to get the branding is via calling that `LoadBranding` func. With `applyDefaults` and a single struct there would be less code, but there's nothing stopping a caller to forget to call it, handing down to a template a struct with non-default values.

I'll keep it as is for now, but we can rework it later if it feels too much.

PR #197, opened 2026-07-08
Fix issues with new update upload feature

doanac:
This is a set of embarrassingly bad bugs I discovered while doing e2e testing against a proper FoundriesFactory and not builds I'd created by hand.

### DIFF
--- a/storage/api/api_storage_tuf.go
+++ b/storage/api/api_storage_tuf.go
@@ -56,6 +56,8 @@ func (s Storage) GenerateTufMeta(tufDir string, opts TargetOptions) error {
 	}
 	if opts.AppVersion != 0 {
 		tgtVer = opts.AppVersion
+	} else {
+		tgtVer++ // increment the target version when no override is provided
 	}
 
 	custom := generatedTargetCustom{
--- a/storage/api/api_storage_tuf.go
+++ b/storage/api/api_storage_tuf.go
@@ -4,10 +4,13 @@
 package api
 
 import (
+	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"time"
+
+	"github.com/secure-systems-lab/go-securesystemslib/cjson"
 
 	"github.com/foundriesio/update-server/clock"
 	"github.com/foundriesio/update-server/context"
@@ -104,11 +107,16 @@ func (s Storage) GenerateTufMeta(tufDir string, opts TargetOptions) error {
 		},
 	}
 
-	targetsMeta, err := s.fs.Tuf.Sign(tuf.RoleTargets, targets.Signed)
+	targetsSig, err := s.fs.Tuf.Sign(tuf.RoleTargets, targets.Signed)
 	if err != nil {
 		return fmt.Errorf("unable to sign targets metadata: %w", err)
 	}
-	targets.Signatures = []tuf.Signature{targetsMeta.Signature}
+	targets.Signatures = []tuf.Signature{targetsSig}
+
+	targetsBytes, targetsMeta, err := marshallMeta(targets, targets.Signed.Version)
+	if err != nil {
+		return fmt.Errorf("unable to marshal targets metadata: %w", err)
+	}
 
 	ss := tuf.AtsTufSnapshot{
 		Signed: tuf.SnapshotMeta{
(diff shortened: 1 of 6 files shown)


storage/api/api_storage_update.go
```diff
@@ -25,7 +25,7 @@ func (s Storage) generateUpdateTuf(updateDir, tag string, overrides TargetOption
 		Tag:     tag,
 	}
 
-	ostreeDir := filepath.Join(updateDir, "ostree")
+	ostreeDir := filepath.Join(updateDir, "ostree_repo")
```
vkhoroz: Please, use this constant: https://github.com/foundriesio/update-server/blob/main/storage/file.go#L64

storage/api/api_storage_update.go
```diff
@@ -93,7 +93,7 @@ func probeOstree(repoPath string, opts *TargetOptions) error {
 		return err
 	}
 
-	if data, err := repo.ReadFile(ref, "/etc/os-release"); err == nil {
+	if data, err := repo.ReadFile(ref, "/usr/lib/os-release"); err == nil {
```
vkhoroz, line 97: Should this be an absolute or a relative path?
IIUC it is concatenated with the root folder, right?
If yes - a relative path is safer to use.

storage/tuf/tuf_crypto.go
```diff
@@ -100,28 +100,12 @@ func (s *Signer) PublicAtsKey() AtsKey {
...
+		return Signature{}, fmt.Errorf("unable to marshal canonical JSON: %w", err)
 	}
 	sig := ed25519.Sign(s.private, msg)
-	sum := sha256.Sum256(msg)
-	return SignedMeta{
-		Signature: Signature{KeyID: s.Id, Method: SigEd25519, Signature: sig},
-		Sha256:    sum[:],
-		Length:    int64(len(msg)),
-	}, nil
+	return Signature{KeyID: s.Id, Method: SigEd25519, Signature: sig}, nil
```
vkhoroz, line 110: I guess by "totally wrong" you mean that hash and length must be for *the entire role*, not *only its signed part*, right?
This is a good catch indeed.

storage/api/api_storage_tuf.go
```diff
@@ -159,32 +172,45 @@ func (s Storage) GenerateTufMeta(tufDir string, opts TargetOptions) error {
...
+// marshall encodes v as canonical JSON and returns its bytes, length, and
+// sha256 hash.
+func marshall(v any) (marshalledBytes, error) {
+	b, err := cjson.EncodeCanonical(v)
+	if err != nil {
+		return marshalledBytes{}, err
+	}
+	sum := sha256.Sum256(b)
+	return marshalledBytes{Bytes: b, Length: int64(len(b)), Sha256: sum[:]}, nil
+}
```
vkhoroz: I kind of understand what this internal structure does, but it is rather weird to see it combine two diverse entities:
1. A marshaled JSON body of the TUF role.
2. A set of attributes needed by another TUF role to compose its body.

I suggest to use two different data structures for these two things.
You may still use one method if you wish or two methods to generate them.

One possible way would be this:
```
func marshalMeta(v any, version int64) ([]byte, tuf.MetaItem, error) {
	// return canonical bytes and a ready to use MetaItem struct.
}
```

There are 3 ways how to work out the version:
1. Pass it, like I did in the example above.
2. Populate it outside the `marshalMeta` function.
3. Define v as an interface which has a method to return version from Signed field (and define it for targets/snapshot/timestamp).

An alternative is to have two separate methods:
```
func marshalMeta(v any) ([]byte, error) {}
func hashMeta(data []byte, version int64) tuf.MetaItem {}
```

storage/api/api_storage_update.go
doanac: Its reading absolute paths from an ostree repo - its not like normal filesystem calls

storage/tuf/tuf_crypto.go
doanac: correct. aklite didn't like what I had done

storage/api/api_storage_tuf.go
doanac: good call. Fixed in force-push

vkhoroz (approved):

PR #199, opened 2026-07-08
tuf: Support importing root keys from a fioctl tarball

doanac:
Add an optional --import-keys flag to the tuf-init command that reads a fioctl/ota-tuf tarball containing the offline root key(s) and root.json. The imported root.json files are stored verbatim, fresh online keys are generated for every role, and a new root.json is created and signed by both the imported (old) root key(s) and the new root key so clients can verify the chain of trust.

Chatting with Copilot on this, it was decided to be more efficient to use our own code for talking with the FIO backend for getting the root.json files rather than trying to integrate with the fioctl client library code.

Fixes bug #189

### DIFF
--- a/cmd/server/tuf_init.go
+++ b/cmd/server/tuf_init.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/foundriesio/update-server/storage"
@@ -49,9 +50,39 @@ func (c TufInitCmd) Run(args CommonArgs) error {
 		if err != nil {
 			return err
 		}
-		return fs.Tuf.ImportTuf(roots, keys)
+		if err := fs.Tuf.ImportTuf(roots, keys); err != nil {
+			return err
+		}
+		printRootKeyBackupNotice(fs)
+		return nil
 	}
-	return fs.Tuf.InitTuf()
+	if err := fs.Tuf.InitTuf(); err != nil {
+		return err
+	}
+	printRootKeyBackupNotice(fs)
+	return nil
+}
+
+// printRootKeyBackupNotice warns the operator that the newly created root key
+// is irreplaceable and must be backed up along with the HMAC secret used to
+// decrypt it.
+func printRootKeyBackupNotice(fs *storage.FsHandle) {
+	rootKeyPath := filepath.Join(fs.Config.TufDir(), "keys", "root.key")
+	hmacPath := filepath.Join(fs.Config.AuthDir(), storage.HmacFile)
+	fmt.Println()
+	fmt.Println("TUF initialization completed successfully.")
+	fmt.Println()
+	fmt.Println("IMPORTANT: A new root key was created at:")
+	fmt.Printf("    %s\n", rootKeyPath)
+	fmt.Println()
+	fmt.Println("This key is encrypted at rest using the HMAC secret at:")
+	fmt.Printf("    %s\n", hmacPath)
+	fmt.Println()
+	fmt.Println("Make a backup copy of BOTH files and store them somewhere safe NOW.")
+	fmt.Println("The root key is useless without the HMAC secret needed to decrypt it.")
+	fmt.Println("If either file is lost it CANNOT be recovered, and you will permanently")
+	fmt.Println("lose the ability to rotate or manage your TUF root of trust.")
+	fmt.Println()
 }
 
 // loadTufKeysArchive opens a fioctl offline keys tarball and extracts the
--- a/cmd/server/tuf_init.go
+++ b/cmd/server/tuf_init.go
@@ -4,15 +4,140 @@
 package main
 
 import (
+	"archive/tar"
+	"compress/gzip"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
 	"github.com/foundriesio/update-server/storage"
+	"github.com/foundriesio/update-server/storage/tuf"
 )
 
-type TufInitCmd struct{}
+// rootJsonSuffix is the suffix ota-tuf uses for root metadata files (e.g.
+// "1.root.json").
+const rootJsonSuffix = ".root.json"
+
+// tufKeyFileSuffix is the suffix fioctl/garage-sign use for private key files
+// stored in the offline keys tarball.
+const tufKeyFileSuffix = ".sec"
+
+type TufInitCmd struct {
+	ImportKeys  string `arg:"--import-keys" help:"Path to a fioctl offline keys tarball with the root key(s) to sign the rotation; enables TUF root import (requires auth-init to have been run first)"`
+	ImportRoots string `arg:"--import-roots" help:"Path to a gzipped tarball containing all root.json files to import"`
+}
 
 func (c TufInitCmd) Run(args CommonArgs) error {
 	fs, err := storage.NewFs(args.DataDir)
 	if err != nil {
 		return err
 	}
+	if c.ImportKeys != "" || c.ImportRoots != "" {
+		if c.ImportKeys == "" {
+			return fmt.Errorf("--import-keys is required to import TUF root metadata")
+		}
+		if c.ImportRoots == "" {
+			return fmt.Errorf("--import-roots is required to import TUF root metadata")
+		}
+		roots, err := loadTufRootsArchive(c.ImportRoots)
+		if err != nil {
+			return err
+		}
+		keys, err := loadTufKeysArchive(c.ImportKeys)
+		if err != nil {
+			return err
+		}
+		return fs.Tuf.ImportTuf(roots, keys)
+	}
 	return fs.Tuf.InitTuf()
+}
+
+// loadTufKeysArchive opens a fioctl offline keys tarball and extracts the
+// root key files it contains. Only private key files (those ending in
+// ".sec") are parsed; all other archive entries are ignored.
+func loadTufKeysArchive(path string) ([]tuf.AtsKey, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("unable to open keys archive: %w", err)
+	}
+	defer f.Close() // nolint:errcheck
+
+	gz, err := gzip.NewReader(f)
+	if err != nil {
+		return nil, fmt.Errorf("unable to open keys archive (expected a gzipped tarball): %w", err)
+	}
+	defer gz.Close() // nolint:errcheck
+
+	var keys []tuf.AtsKey
+	tr := tar.NewReader(gz)
+	for {
+		hdr, err := tr.Next()
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			return nil, fmt.Errorf("unable to read keys archive: %w", err)
+		}
+		if hdr.Typeflag == tar.TypeDir || !strings.HasSuffix(hdr.Name, tufKeyFileSuffix) {
+			continue
+		}
+
+		if data, err := io.ReadAll(tr); err != nil {
+			return nil, fmt.Errorf("unable to read %s from keys archive: %w", hdr.Name, err)
+		} else {
+			var key tuf.AtsKey
+			if err := json.Unmarshal(data, &key); err != nil {
+				return nil, fmt.Errorf("unable to parse key file %s: %w", hdr.Name, err)
+			}
+			if key.KeyType != "" && key.KeyValue.Private != "" {
+				keys = append(keys, key)
+			}
+		}
+	}
+	if len(keys) == 0 {
+		return nil, fmt.Errorf("keys archive does not contain any valid private key files (expected .sec files with private key material)")
+	}
+	return keys, nil
+}
+
+// loadTufRootsArchive reads a gzipped tar archive and returns the raw bytes of
+// every root.json file it contains. Only files ending in ".root.json" are
+// extracted; all other archive entries are ignored.
+func loadTufRootsArchive(path string) ([][]byte, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("unable to open roots archive: %w", err)
+	}
+	defer f.Close() // nolint:errcheck
+
+	gz, err := gzip.NewReader(f)
+	if err != nil {
+		return nil, fmt.Errorf("unable to open roots archive (expected a gzipped tarball): %w", err)
+	}
+	defer gz.Close() // nolint:errcheck
+
+	var roots [][]byte
+	tr := tar.NewReader(gz)
+	for {
+		hdr, err := tr.Next()
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			return nil, fmt.Errorf("unable to read roots archive: %w", err)
+		}
+		if hdr.Typeflag == tar.TypeDir || !strings.HasSuffix(hdr.Name, rootJsonSuffix) {
+			continue
+		}
+
+		if data, err := io.ReadAll(tr); err != nil {
+			return nil, fmt.Errorf("unable to read %s from roots archive: %w", hdr.Name, err)
+		} else {
+			roots = append(roots, data)
+		}
+	}
+	if len(roots) == 0 {
+		return nil, fmt.Errorf("roots archive does not contain any root.json files (expected files ending in %q)", rootJsonSuffix)
+	}
+	return roots, nil
 }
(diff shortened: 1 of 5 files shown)


storage/file_tuf_import.go
```diff
@@ -0,0 +1,247 @@
...
+			break
+		} else if err != nil {
+			return nil, fmt.Errorf("unable to read keys archive: %w", err)
+		}
+		if hdr.Typeflag == tar.TypeDir || !strings.HasSuffix(hdr.Name, tufKeyFileSuffix) {
+			continue
+		}
+
+		var buf bytes.Buffer
+		if _, err := io.Copy(&buf, io.LimitReader(tr, maxImportFileSize+1)); err != nil {
```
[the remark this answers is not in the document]
doanac: The code is using an io.LimitReader which provides a better error an than io.CopyN

cmd/server/tuf_init.go
```diff
@@ -4,15 +4,51 @@
...
+	ImportKeys    string `arg:"--import-keys" help:"Path to a fioctl offline keys tarball with the root key(s) to sign the rotation; enables TUF root import (requires auth-init to have been run first)"`
+	ImportConfig  string `arg:"--import-config" help:"Path to a fioctl.yaml used to authenticate to the Foundries.io API (default: $FIOCTL_CONFIG or ~/.config/fioctl.yaml)"`
+	ImportFactory string `arg:"--import-factory" help:"Foundries.io factory whose TUF root metadata should be imported; defaults to the factory in the fioctl config"`
+}
+
+// TODO - factories with ci/prod roots - how could we serve something for both?
+// maybe it doesn't matter? If a device is already running with a root, it will have the root
+// json it needs and n+1 will be signed properly.
+// If a device is new, it will get the new root from the server - it can get whatever
+// history we choose and it will be okay.
```
vkhoroz: A production TUF root is signed by enough keys to satisfy both (CI only requires one signature, while prod requires one or more).
Thus, using the production TUF root for the starter (if factories have both) makes more sense.
A CI TUF root might not work for production devices which have prod root preloaded with signature threshold > 1.
After we generated a new root with new keys, we're settled.

cmd/server/tuf_config.go
```diff
@@ -0,0 +1,115 @@
...
+	Token             string            `yaml:"token"`
+	ClientCredentials fioctlOAuth       `yaml:"clientcredentials"`
+	ExtraHeaders      map[string]string `yaml:"extraheaders"`
+	Server            fioctlServer      `yaml:"server"`
+}
+
+// defaultFioctlConfigPath returns the location fioctl reads its config from,
+// honouring the FIOCTL_CONFIG environment variable and falling back to
+// ~/.config/fioctl.yaml.
+func defaultFioctlConfigPath() (string, error) {
```
vkhoroz: I'm not very fond of us having to copy a load of Fioctl logic into update-server, although I understand that necessity. I'm not fond of having to talk to ota-lite too.

Is it possible to instead have a hidden command in fioctl which downloads all necessary root.json files into some folder (better a tarball), and then tuf-init would take that tarball and import it?
Users who used Foundries.io would know how to use this; and users new to update-server don't even need to know it.
Furthermore, if they know the (pretty simple) structure of that tarball - they'd be able to import roots from other products.

vkhoroz: One more thing to consider - those who did not take their TUF root offline yet (many of our existing customers).
IIUC we need to download to initial root key for them too, right?

vkhoroz: Generalizing my comment, what I think would be a better alternative is running these two commands:
```
# Downloads all roots and packs them into roots.tar.gz
# together with the currently active root keys from root-keys.tar.gz.
# If there are production roots - take them; otherwise take CI roots.
# If user did not yet take their TUF root key offline - they pass --initial to download it.
fioctl keys tuf export-roots <roots.tar.gz> <root-keys.tar.gz>|--initial

# Imports existing roots, and then generates new set of keys, sets signature thresholds to 1,
# and signs a new root with old/new keys to meet the previous threshold.
update-server init-tuf --import-roots <roots.tar.gz>
```

doanac: We can build a new version of fioctl with this command. I'm not sure what the path of least resistance for the user would be. I'm happy to do it this way and hold off on accessing the API until someone asks for it.

> IIUC we need to download to initial root key for them too, right?

Correct. I think all our customers that would want to migrate have already done this

cmd/server/tuf_init.go
doanac: the only way this won't work is if someone had signed root.json with two root keys. Nobody has done that. The number of target sigs is irrelevant when generating a new root.json. The only requirement is that, if using a new key for the new version, you also sign it with the old key.

doanac: I forgot to remove the comment - i wrote this while testing.

vkhoroz: If none of our customers (willing to migrate) have ever set root signature threshold to more than 1 - then we can use any root, and CI is simply easier to do this.

One subtle thing to check: they should have never set root signature threshold to 2 (have no devices with such root preloaded); it is not enough to have a current threshold set to 1.

doanac: good call. I addressed this in a force push by rejecting a request to import if root.Signe.Roles[root].Threshold > 1

cmd/server/tuf_config.go
doanac: The force push removed the fioctl api logic and requires a tarball instead. I'll test this e2e tomorrow after implementing this in fioctl

cmd/server/tuf_init.go
```diff
@@ -4,15 +4,92 @@
...
+			break
+		} else if err != nil {
+			return nil, fmt.Errorf("unable to read roots archive: %w", err)
+		}
+		if hdr.Typeflag == tar.TypeDir || !strings.HasSuffix(hdr.Name, rootJsonSuffix) {
+			continue
+		}
+
+		var buf bytes.Buffer
+		if _, err := io.Copy(&buf, io.LimitReader(tr, maxRootFileSize+1)); err != nil {
```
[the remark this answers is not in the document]
doanac: I'm using a limit reader.

cmd/server/tuf_init.go
```diff
@@ -4,15 +4,92 @@
 package main
```
vkhoroz, line 4: The commit message is no longer relevant in a part about talking to the API.
We've decoupled off that dependency via the new Fioctl command.

cmd/server/tuf_init.go
```diff
@@ -4,15 +4,92 @@
...
+	"strings"
+
 	"github.com/foundriesio/update-server/storage"
 )
 
-type TufInitCmd struct{}
+// maxRootFileSize caps the size of a single decompressed root.json read from
+// the roots archive. It guards against decompression bombs while remaining far
+// larger than any legitimate root.json file.
+const maxRootFileSize = 16 << 20 // 16 MiB
```
vkhoroz: This is literally infinity.  I'm not sure we need to be trying to appease some mechanistic AI validator "just in case", and get our code more complex.

We're not downloading some nonsense from the web; a user supplies an input tarball to us explicitly, and they can see its size cleanly.
If they plan to break themselves - they'll find a way regardless of how hard we try to protect, as they are a god at their own laptop.

storage/file_tuf_import.go
```diff
@@ -0,0 +1,255 @@
...
+	"time"
+
+	"github.com/foundriesio/update-server/clock"
+	"github.com/foundriesio/update-server/storage/tuf"
+)
+
+// maxImportFileSize caps the size of a single decompressed file read from the
+// keys archive. It guards against decompression bombs while remaining far
+// larger than any legitimate key file.
+const maxImportFileSize = 16 << 20 // 16 MiB
```
vkhoroz: The same constant in a different place (the other one was in `tuf_init.go`); which is anyway useless.

storage/file_tuf_import.go
```diff
@@ -0,0 +1,255 @@
...
+// a fioctl offline keys tarball containing the offline private key(s) for the
+// root role. Every imported
+// root.json is stored verbatim, fresh online keys are generated for every role,
+// and a new root.json (version = highest imported version + 1) is created and
+// signed by both the imported (old) root key(s) and the newly generated root
+// key so that clients can verify the chain of trust from the previously
+// trusted root.
+//
+// It fails if TUF data already exists.
+func (h TufFsHandle) ImportTuf(rootJSONs [][]byte, archivePath string) error {
```
vkhoroz: 1. The function signature doesn't seem consistent: we pass roots as parsed files and keys as a path to an archive.
    If there is no hidden value to keep the root archive parsing logic inside `tuf_init.go`, maybe, we pass a path to both archives into this function, and make that roots parsing internal too?

2. This function imports roots, not a generic TUF archive. Also, a TUF is already a part of the class name. So, maybe `ImportRoots`?

3. A better name for the second variable is smth like `keysArchivePath` or simply `keysPath`.

storage/file_tuf_import.go
```diff
@@ -0,0 +1,255 @@
...
+		if err := json.Unmarshal(raw, &root); err != nil {
+			return fmt.Errorf("unable to parse root metadata: %w", err)
+		}
+		if len(root.Signed.Roles) == 0 {
+			return fmt.Errorf("provided metadata does not look like a TUF root.json")
+		}
+		roots = append(roots, importedRoot{raw: slices.Clone(raw), root: root})
+	}
+
+	f, err := os.Open(archivePath)
```
vkhoroz: So, IIUC this keys path can be anything, it is not relative to the TufFsHandle.root, right?
That's totally fine, I'm just asking.

storage/file_tuf_import.go
```diff
@@ -0,0 +1,255 @@
...
+	roots := make([]importedRoot, 0, len(rootJSONs))
+	for _, raw := range rootJSONs {
+		var root tuf.AtsTufRoot
+		if err := json.Unmarshal(raw, &root); err != nil {
+			return fmt.Errorf("unable to parse root metadata: %w", err)
+		}
+		if len(root.Signed.Roles) == 0 {
+			return fmt.Errorf("provided metadata does not look like a TUF root.json")
+		}
+		roots = append(roots, importedRoot{raw: slices.Clone(raw), root: root})
```
vkhoroz: There is some subtle magic in here: why do we need to clone the `raw` slice?

storage/file_tuf_import.go
```diff
@@ -0,0 +1,255 @@
...
+		return fmt.Errorf("unable to read HMAC secret (run auth-init first): %w", err)
+	} else if len(hmacSecret) == 0 {
+		return fmt.Errorf("HMAC secret is empty; run auth-init first")
+	}
+
+	// The highest-version imported root is the trust anchor to chain from.
+	slices.SortFunc(roots, func(a, b importedRoot) int {
+		return a.root.Signed.Version - b.root.Signed.Version
+	})
+	base := roots[len(roots)-1]
```
vkhoroz, line 75: I don't know where to put this comment; but I think this place, after we parsed and sorted all roots in the most appropriate.

So, I guess we should validate that imported roots are non-interrupted, i.e. all natural numbers between `1..base` are included. Otherwise, the TUF roots chain of trust is broken.

storage/file_tuf_import.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+	signatures := []tuf.Signature{newSigned}
+	for _, old := range oldSigners {
+		oldSigned, err := old.Sign(newRoot.Signed)
+		if err != nil {
+			return fmt.Errorf("unable to sign new root metadata with imported key %s: %w", old.Id, err)
+		}
+		signatures = append(signatures, oldSigned)
+	}
+	newRoot.Signatures = signatures
```
vkhoroz: It does not need to be signed by all old root keys.
It must be signed by one new key and one old key, as long as the signature threshold is 1 (we're checking for that above).

storage/file_tuf_import.go
```diff
@@ -0,0 +1,255 @@
...
+	var signers []*tuf.ImportSigner
+	for _, key := range candidateKeys {
+		signer, err := tuf.ImportSignerFromAtsKey(key)
+		if err != nil {
+			// Not a usable signing key (e.g. public-only or unsupported type).
+			continue
+		}
+		if wanted[signer.Id] {
+			signers = append(signers, signer)
+			delete(wanted, signer.Id)
```
vkhoroz: I guess this is to prevent duplicates; but it looks weird, especially having that no duplicates are possible inside key IDs in practice.

vkhoroz (commented):
This looks good to me.

I'm only a bit surprised by a sheer amount of code needed to do this.
But that's irrelevant. You did a great job.

cmd/server/tuf_init.go
doanac: fair point. The original logic was grabbing json from REST APIs. The AI reviewer still thinks its not safe because it doesn't understant io.LimitReader. I'll drop this.

storage/file_tuf_import.go
doanac: Its not just a check for duplicates, it also keeps a list of what private keys were missing. So, `delete(wanted..` is needed for a quality error message. I tried to refactoring a bit but it made things worse

storage/file_tuf_import.go
doanac: I updated the message in the force push

storage/file_tuf_import.go
doanac: good catch. updated in force-push

storage/file_tuf_import.go
doanac: AI slop. I removed it.

storage/file_tuf_import.go
doanac: its not. I refactored this logic to do all the file reading in the subcommand rather than in the API layer

storage/file_tuf_import.go
doanac: I refactored things in the force push to read the file content in the subcommand and then pass what we've parsed to the api.

storage/file_tuf_import.go
vkhoroz: The problem was not with the message but with the logic - it was signing TUF root with every old signing key.

In practice, however, there is always only one old key, as none of our users used multiple root keys yet IIUC.
(Even if they were using multiple keys, that flow recommends to keep only one key per tarball... i.e. there is never more than 1 *matching* key, even if root.json lists multiple keys).
But, this code makes a false sense of multiple old signatures being added.
There is no harm in that, but no good either.

What if we change `matchRootSigners` to only return the first signer it found?
That would put in stone exactly what we are actually doing.

storage/file_tuf_import.go
vkhoroz: I don't see any error message for the "missing private keys".
This `wanted` map is never used outside this function.

cmd/server/tuf_init.go
```diff
@@ -49,9 +50,39 @@ func (c TufInitCmd) Run(args CommonArgs) error {
...
+	fmt.Printf("    %s\n", rootKeyPath)
+	fmt.Println()
+	fmt.Println("This key is encrypted at rest using the HMAC secret at:")
+	fmt.Printf("    %s\n", hmacPath)
+	fmt.Println()
+	fmt.Println("Make a backup copy of BOTH files and store them somewhere safe NOW.")
+	fmt.Println("The root key is useless without the HMAC secret needed to decrypt it.")
+	fmt.Println("If either file is lost it CANNOT be recovered, and you will permanently")
+	fmt.Println("lose the ability to rotate or manage your TUF root of trust.")
+	fmt.Println()
```
vkhoroz, line 85: Maybe, use a multi-line format, like this:
```suggestion
	fmt.Printf(`
TUF initialization completed successfully.

IMPORTANT: A new root key was created at:
	%s

This key is encrypted at rest using the HMAC secret at:
	%s

Make a backup copy of BOTH files and store them somewhere safe NOW.
The root key is useless without the HMAC secret needed to decrypt it.
If either file is lost it CANNOT be recovered, and you will permanently
lose the ability to rotate or manage your TUF root of trust.
`,
	rootKeyPath, hmacPath)
```

vkhoroz (approved):
LGTM

I'd still recommend to sort out that feeling of adding multiple old key signatures.
(That what the code seems to do, but we only ever have one matching old key in the tarball).

Feel free to merge when you like it.

storage/file_tuf_import.go
doanac: right below if `len(signers) == 0` - we print what keyIDs were missing. However, if we change this to just be "findFirstSigner" like you are talking about, then things get easier

vkhoroz: fwiw when we append at least once - that message is never printed, regardless if we have this `delete(...)` here or not.

PR #202, opened 2026-07-10
Improve the way we create a base page context

doanac:
We've had this split where the auth package and UI package have their own logic for creating this base page context.

Every time we've added a new field, this time its branding stuff, we've broken some page's rendering logic. This change introduces logic to consolidate things and fights circular dependencies.

There are even better ways this could be done. However, this is the least invasive way and probably serves the size of our code well enough to survive with for the foreseeable future.

### DIFF
--- a/server/ui/web/handlers.go
+++ b/server/ui/web/handlers.go
@@ -18,9 +18,9 @@ import (
 
 	"github.com/foundriesio/update-server/auth"
 	"github.com/foundriesio/update-server/server"
+	"github.com/foundriesio/update-server/server/ui/pagectx"
 	"github.com/foundriesio/update-server/server/ui/web/templates"
 	"github.com/foundriesio/update-server/storage/users"
-	"github.com/foundriesio/update-server/version"
 )
 
 type handlers struct {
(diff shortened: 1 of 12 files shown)


server/ui/web/handlers.go
```diff
@@ -95,34 +97,10 @@ func RegisterHandlers(e *echo.Echo, storage *users.Storage, authProvider auth.Pr
...
-	return baseCtx{
-		User:      CtxGetSession(c.Request().Context()).User,
-		Title:     title,
-		BrandName: h.branding.Title,
-		LogoPath:  logoPath,
-		NavItems:  h.genNavItems(selected),
-		CsrfToken: csrfToken,
-		Version:   version.Version,
-	}
+	return h.pages.Base(c, title, selected)
```
vkhoroz, line 103: Just to clarify on the commit message: we were usually breaking the auth pages, not literally "every page" when introducing new base context fields, right?

vkhoroz (approved):
LGTM

server/ui/web/handlers.go
doanac: that's correct

PR #203, opened 2026-07-13
Perf test idea

StealthyCoder:
## What

Implements a self-contained, fast `contrib/perf-test/` Locust harness for
`dg-sat`, replacing `contrib/scale-test`. Registers/drives thousands of
simulated devices over mTLS, plus an admin actor over the REST API, plus a
TUF check-for-update/download flow — with isolated, tag-selectable
scenarios rather than one monolithic mixed run.

## Why a new folder instead of extending `scale-test`

`contrib/scale-test`'s cert generation forks `openssl` per device — at
5000 devices that's tens of thousands of process spawns, minutes just to
get to a runnable state. `contrib/perf-test/gen-certs` is a single Go
binary doing all keygen+signing in-process: 5000 ECDSA P-256 device certs
in ~250ms. Everything else (setup, seeding, the Locust file itself) is
also self-contained under one `docker compose`/`Makefile` entry point.
`scale-test` is removed in this PR rather than kept alongside it — two
parallel, easily-diverging perf-testing setups isn't worth the upkeep.

## What it covers

| Actor | Endpoints |
|---|---|
| Device (mTLS `:8443`) | `GET /device`, `GET /config`, `POST /events`, `GET /repo/timestamp.json`\|`snapshot.json`\|`targets.json`, `POST /ostree/download-urls`, `GET /ostree/config` |
| Admin (REST `:8080`) | `GET /v1/devices` (paginated) |

This is a subset of the server's full HTTP surface by design — see
"Follow-up" below.

## Design notes

- **Fixture seeding is at the storage layer, not through the REST upload
  API.** `gen-certs --seed-update` writes an unsigned-but-structurally-valid
  `targets.json`/`snapshot.json`/`timestamp.json`/root, plus a real 256KiB
  ostree object, and assigns it via a rollout — all before `dg-sat serve`
  ever starts. The gateway doesn't verify TUF signatures on `GET` (that's a
  TUF-client concern), so unsigned content is legitimate here and far
  cheaper than driving a real signed tar upload through the REST API for
  zero benefit on the paths this harness actually exercises.
- **Rollouts are assigned by UUID, not group.** A device that self-registers
  via real mTLS always has `group_name=""` (`DeviceCreate` never sets it) —
  a rollout keyed by `groups` would silently match nothing. Confirmed by
  reading `storage/gateway/storage.go` before relying on it.
- **Seeded devices get their real extracted pubkey, not a placeholder.** A
  fake key would make the real mTLS handshake hit "key rotation not
  supported" on first contact, since the DB pubkey wouldn't match the cert
  `gen-certs` already wrote to disk for that device.
- **Admin traffic uses absolute URLs, not `self.host`.** Locust's runner
  overwrites every registered `User` class's `host` with the mTLS device
  host immediately before each spawn (`runners.py`, confirmed by reading
  Locust 2.43.4's source, not assumed from docs) — `AdminUser` can't rely
  on `HttpSession.base_url` and builds full URLs instead, including when
  following the paginated `Link` header (server returns it as a relative
  path, so it must be re-anchored per page).
- **Every task is tagged** (`device:*`/`admin:*`/`update:*`,
  `update:check`/`update:download`), mirroring the `@locust.tag(...)`
  convention already used elsewhere for Locust harnesses in this org, so
  `--tags`/`--exclude-tags` can isolate one scenario at a time —
  `ceteris paribus`, without steady-state/admin/update traffic all
  competing for the same SQLite writer in the same run. This replaced an
  earlier custom `--update-flow-weight` CLI flag that only gated the whole
  flow on/off; native tag filtering is strictly more capable and one less
  thing to maintain.
- **Named `make locust-*` targets + `PROFILE=`/`SCENE=` composition.**
  `make locust-update-check`, `locust-update`, `locust-admin`,
  `locust-steady-state` cover the common cases without needing to remember
  the `--tags` incantation; `profiles/*.mk`/`scenes/*.mk` let scale and
  task-selection presets be composed (`make headless-scenario PROFILE=smoke
  SCENE=update-check`), reusable by name, still overridable on the command
  line (command-line assignments win over both target-specific and
  included-file assignments — verified against GNU Make's actual
  precedence rules, not assumed).
- **`make help` / bare `make`** lists every target from the `##`
  doc-comments already on them, plus whatever's actually on disk under
  `profiles/`/`scenes/` — can't drift out of sync with what's implemented,
  unlike a hand-maintained list would.
- **`DRY_RUN=1`** prints the fully-resolved `docker compose` command instead
  of running it — narrower than `make -n`, which walks the whole
  prerequisite chain and echoes every target's recipe along the way.

## ⚠️ Known limitation: spawn-rate-dependent registration latency

Under a device cold-start burst, `dg-sat`'s single-writer SQLite
(`db.SetMaxOpenConns(1)`) serializes first-contact writes; a small number
of unlucky requests queue before finally succeeding (e.g. ~240s tail at
our 80/s spawn cap, 5000 devices). Zero failures either way — it's a
latency problem, not a correctness one.

A follow-up spawn-rate scan (2000 devices, steady-state traffic only,
rates 10/20/40/80/160) found the tail has a clear knee: median/p95 latency
stay flat (10ms/~25ms) at every rate tested, but max latency grows
roughly linearly from 10→20/s then jumps nonlinearly from 20/s onward
(741ms → 1.4s → 6.9s → 26.5s → 39.8s), still with zero failures at every
rate including 160/s. The `Makefile`'s 80/s default sits well past that
knee. Full methodology, caveats, and reproduction steps in
`contrib/perf-test/baselines/ratescan-2026-07-13.md`; root-cause writeup
in `contrib/perf-test/baselines/baseline-2026-07-11.md`.

## Follow-up

The harness currently covers a fraction of the server's REST surface
(missing: apps/registry download, device hardware/network/apps-state
reporting, fio-tests, and the whole admin configs/rollouts/TUF-management
domain). Planned as a sequence of small follow-up PRs — one domain at a
time, each waiting on review of the previous before starting — rather than
one large diff. Tracked separately, not part of this PR's diff.

## Testing

- Full 5000-device runs at both ends of this PR (unseeded baseline, and
  seeded + update-flow + admin mixed run): 0 failures, 0 exceptions, full
  device registration confirmed via the `Link` header's `last` offset.
  Results and analysis in `contrib/perf-test/baselines/`.
- Every named `make` target and `PROFILE=`/`SCENE=` combination
  dry-run-verified (`make -n`) and then actually run against a live
  seeded server at small scale before being trusted at full scale.
- `make headless` / `make run` (no extra flags) re-verified unaffected
  after each change that touched shared plumbing (config resolution,
  tagging, the Makefile rewrite) — regression-checked, not just assumed
  from reading the diff.


contrib/perf-test/gen-certs/main.go
```diff
@@ -0,0 +1,190 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
doanac, line 1: When you move to the latest main, you'll see this: https://github.com/foundriesio/update-server/tree/main/cmd/seed I'm not sure if its worth consolidating code - but a single command to generate proper device certs isn't a bad idea. A customer could want this for manufacturing

PR #205, opened 2026-07-14
Add support for setting up user PKI from scratch

doanac:
This feature helps users w/o the intial PKI created from the FoundriesFactory product

Fixes #173

### DIFF
--- a/cmd/server/pki_init.go
+++ b/cmd/server/pki_init.go
@@ -1,0 +1,152 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+package main
+
+import (
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"fmt"
+	"time"
+
+	"github.com/foundriesio/update-server/storage"
+)
+
+type PkiInitCmd struct {
+	DnsName       string `arg:"required" help:"DNS host name devices address this gateway with"`
+	Factory       string `arg:"required"`
+	TlsExpiryDays int    `default:"365" help:"TLS certificate validity in days"`
+	CaExpiryDays  int    `default:"7300" help:"Root and device CA certificate validity in days"`
+}
+
+func (c PkiInitCmd) Run(args CommonArgs) error {
+	fs, err := storage.NewFs(args.DataDir)
+	if err != nil {
+		return err
+	}
+	if err = fs.Certs.AssertCleanPki(); err != nil {
+		return err
+	}
+
+	// 1. Root CA
+	rootKey, err := generateEcKey()
+	if err != nil {
+		return err
+	}
+	serial, err := newSerial()
+	if err != nil {
+		return err
+	}
+	rootTemplate := &x509.Certificate{
+		Subject: pkix.Name{
+			CommonName:         c.Factory + "-root",
+			OrganizationalUnit: []string{c.Factory},
+		},
+		SerialNumber:          serial,
+		NotBefore:             time.Now(),
+		NotAfter:              time.Now().AddDate(0, 0, c.CaExpiryDays),
+		IsCA:                  true,
+		KeyUsage:              x509.KeyUsageCertSign | x509.KeyUsageCRLSign,
+		BasicConstraintsValid: true,
+	}
+	rootPem, err := signCert(rootTemplate, rootTemplate, &rootKey.PublicKey, rootKey)
+	if err != nil {
+		return fmt.Errorf("error creating root CA: %w", err)
+	}
+	// signCert produced a self-signed cert; parse it back so it can be used
+	// to issue child certificates.
+	rootBlock, _ := pem.Decode(rootPem)
+	if rootBlock == nil {
+		return fmt.Errorf("unexpected error decoding root CA PEM")
+	}
+	rootCrt, err := x509.ParseCertificate(rootBlock.Bytes)
+	if err != nil {
+		return fmt.Errorf("error parsing root CA: %w", err)
+	}
+	rootKeyPem, err := marshalKeyPem(rootKey)
+	if err != nil {
+		return err
+	}
+	if err := fs.Certs.WriteFile(storage.CertsRootKeyFile, rootKeyPem); err != nil {
+		return fmt.Errorf("unable to store root CA key: %w", err)
+	}
+	if err := fs.Certs.WriteFile(storage.CertsRootPemFile, rootPem); err != nil {
+		return fmt.Errorf("unable to store root CA cert: %w", err)
+	}
+
+	// 2. TLS keypair signed by the root CA (reusing the create-csr/sign-csr path)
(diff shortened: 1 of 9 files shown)


cmd/server/pki_init.go
```diff
@@ -0,0 +1,152 @@
...
+	// signCert produced a self-signed cert; parse it back so it can be used
+	// to issue child certificates.
+	rootBlock, _ := pem.Decode(rootPem)
+	if rootBlock == nil {
+		return fmt.Errorf("unexpected error decoding root CA PEM")
+	}
+	rootCrt, err := x509.ParseCertificate(rootBlock.Bytes)
+	if err != nil {
+		return fmt.Errorf("error parsing root CA: %w", err)
+	}
```
vkhoroz, line 66: This deserves one more utility function like e.g. `parseCertPem`. I guess it is used in several places.

cmd/server/pki_init.go
```diff
@@ -0,0 +1,152 @@
...
+	"fmt"
+	"time"
+
+	"github.com/foundriesio/update-server/storage"
+)
+
+type PkiInitCmd struct {
+	DnsName      string `arg:"required" help:"DNS host name devices address this gateway with"`
+	Factory      string `arg:"required"`
+	ExpiryDays   int    `default:"365" help:"TLS certificate validity in days"`
```
vkhoroz: Can you rename this to `TlsExpiryDays`?

(I've already wrote a comment below before scrolling back to see that my comment was wrong.)

docs/quick-start.md
```diff
@@ -16,7 +16,38 @@ For Linux and Mac, make sure to `chmod +x fioserver`.
 
 ## Configure Mutual TLS
 
-### Creat Certificate Signing requests for TLS
+There are two ways to set up the mutual TLS certificates devices use to
+authenticate with this server:
```
vkhoroz: I had to read that `certificates devices` part 3 times to understand what's going on.
Maybe,
```suggestion
There are two ways to set up the mutual TLS certificates which are used by devices to
authenticate with this server:
```

docs/quick-start.md
```diff
@@ -16,7 +16,38 @@ For Linux and Mac, make sure to `chmod +x fioserver`.
...
+
+* **Bootstrap a new PKI from scratch** with `pki-init` — use this if you do
+  not already have a FoundriesFactory PKI (see below).
```
vkhoroz, line 23: I'd also add that once the PKI was bootstrapped, the user can generate more device CAs or rotate TLS cert using the commands below (the other option).

vkhoroz (approved):
LGTM

This is awesome.

docs/quick-start.md
doanac: I have that on the roadmap: https://github.com/foundriesio/update-server/issues/204

cmd/server/pki_init.go
doanac: I've merged w/o this because I'm not sure the solution is much better than having some repeat code. Here's one idea I have if we want to delete a couple lines of code:
```
diff --git a/cmd/server/csr_sign.go b/cmd/server/csr_sign.go
index 62d72f05..743729f9 100644
--- a/cmd/server/csr_sign.go
+++ b/cmd/server/csr_sign.go
@@ -64,44 +64,38 @@ func (c CsrSignCmd) Run(args CommonArgs) error {
        return nil
 }
 
-func loadCert(path string) (*x509.Certificate, error) {
-       cert, err := os.ReadFile(path)
-       if err != nil {
-               return nil, fmt.Errorf("unable to read cert: %w", err)
-       }
-
-       first, rest := pem.Decode(cert)
+func pemBytesToObject[T any](pemBytes []byte, parse func([]byte) (T, error)) (T, error) {
+       first, rest := pem.Decode(pemBytes)
        if first == nil || len(strings.TrimSpace(string(rest))) > 0 {
-               return nil, fmt.Errorf("malformed PEM data for %s", path)
+               var zero T
+               return zero, fmt.Errorf("malformed PEM data")
        }
 
-       return x509.ParseCertificate(first.Bytes)
+       return parse(first.Bytes)
 }
 
-func loadCsr(path string) (*x509.CertificateRequest, error) {
-       csr, err := os.ReadFile(path)
+func loadPemFile[T any](path string, parse func([]byte) (T, error)) (T, error) {
+       pemBytes, err := os.ReadFile(path)
        if err != nil {
-               return nil, fmt.Errorf("unable to read csr: %w", err)
+               var zero T
+               return zero, fmt.Errorf("unable to read %s: %w", path, err)
        }
-
-       first, rest := pem.Decode(csr)
-       if first == nil || len(strings.TrimSpace(string(rest))) > 0 {
-               return nil, fmt.Errorf("malformed PEM data for %s", path)
+       obj, err := pemBytesToObject(pemBytes, parse)
+       if err != nil {
+               var zero T
+               return zero, fmt.Errorf("%w for %s", err, path)
        }
-
-       return x509.ParseCertificateRequest(first.Bytes)
+       return obj, nil
 }
 
-func loadKey(path string) (*ecdsa.PrivateKey, error) {
-       key, err := os.ReadFile(path)
-       if err != nil {
-               return nil, fmt.Errorf("unable to read pkey: %w", err)
-       }
+func loadCert(path string) (*x509.Certificate, error) {
+       return loadPemFile(path, x509.ParseCertificate)
+}
 
-       first, rest := pem.Decode(key)
-       if first == nil || len(strings.TrimSpace(string(rest))) > 0 {
-               return nil, fmt.Errorf("malformed PEM data for %s", path)
-       }
+func loadCsr(path string) (*x509.CertificateRequest, error) {
+       return loadPemFile(path, x509.ParseCertificateRequest)
+}
 
-       return x509.ParseECPrivateKey(first.Bytes)
+func loadKey(path string) (*ecdsa.PrivateKey, error) {
+       return loadPemFile(path, x509.ParseECPrivateKey)
 }
diff --git a/cmd/server/pki_init.go b/cmd/server/pki_init.go
index cd0846ed..7a3cc5cd 100644
--- a/cmd/server/pki_init.go
+++ b/cmd/server/pki_init.go
@@ -6,7 +6,6 @@ package main
 import (
        "crypto/x509"
        "crypto/x509/pkix"
-       "encoding/pem"
        "fmt"
        "time"
 
@@ -56,11 +55,7 @@ func (c PkiInitCmd) Run(args CommonArgs) error {
        }
        // signCert produced a self-signed cert; parse it back so it can be used
        // to issue child certificates.
-       rootBlock, _ := pem.Decode(rootPem)
-       if rootBlock == nil {
-               return fmt.Errorf("unexpected error decoding root CA PEM")
-       }
-       rootCrt, err := x509.ParseCertificate(rootBlock.Bytes)
+       rootCrt, err := pemBytesToObject(rootPem, x509.ParseCertificate)
        if err != nil {
                return fmt.Errorf("error parsing root CA: %w", err)
        }
```

vkhoroz: So, you needed a generic, because we might have two different objects.

It might seem complex, but it is quite good; and reusing this code might prevent silly bugs.

PR #206, opened 2026-07-15
storage: Touch rollouts.log when creating an update

doanac:
We have this annoying bug where you create:
 * Create update (no devices have started to update)
 * Tail the rollout.log - get error

This creates an empty file so the tailing logic will work as soon as the update is created.

### DIFF
--- a/storage/api/api_storage.go
+++ b/storage/api/api_storage.go
@@ -554,7 +554,12 @@ func (s Storage) CreateUpdate(tag, updateName, uploadedBy string, opts TargetOpt
 	if err != nil {
 		return err
 	}
-	return s.stmtUpdateInsert.run(tag, updateName, uploadedBy)
+	if err := s.stmtUpdateInsert.run(tag, updateName, uploadedBy); err != nil {
+		return err
+	}
+	// Create an empty file so that users don't get errors trying to tail the update/rollout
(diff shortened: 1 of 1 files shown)


storage/api/api_storage.go
```diff
@@ -554,6 +554,8 @@ func (s Storage) CreateUpdate(tag, updateName, uploadedBy string, opts TargetOpt
 	if err != nil {
 		return err
 	}
+	// Create an empty file so that users don't get errors trying to tail the update/rollout
+	_ = s.fs.Updates.Logs.AppendFile(tag, updateName, storage.LogRolloutsFile, "")
 	return s.stmtUpdateInsert.run(tag, updateName, uploadedBy)
```
vkhoroz: I think we should insert and then update. Otherwise, if the insert fails, we might end up with a stale file.
```suggestion
	err = s.stmtUpdateInsert.run(tag, updateName, uploadedBy)
	if err != nil {
		return err
	}
	// Create an empty file so that users don't get errors trying to tail the update/rollout
	_ = s.fs.Updates.Logs.AppendFile(tag, updateName, storage.LogRolloutsFile, "")
	return nil
```

vkhoroz (approved):

storage/api/api_storage.go
doanac: fixed in force push

PR #210, opened 2026-07-15
Add support for missing hardware-id in app only updates

doanac:
Fixes issue #209

### DIFF
--- a/server/ui/web/templates/updates.html
+++ b/server/ui/web/templates/updates.html
@@ -52,6 +52,9 @@
             <label for="upload-target-name">Target name</label>
             <input type="text" id="upload-target-name" placeholder="Override the target name" />
 
+            <label for="upload-hardware-id">Hardware ID</label>
+            <input type="text" id="upload-hardware-id" placeholder="Override the hardware id" />
+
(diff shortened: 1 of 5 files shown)


server/ui/web/templates/updates.html
```diff
@@ -243,6 +243,22 @@ <h2>{{.Title}}</h2>
...
+      // hardware id and target name cannot be derived from the update content
+      // and must be supplied via the advanced options.
+      const hasOstreeTree = Array.from(files).some((file) => {
+        const rel = file.webkitRelativePath || '';
+        return rel.split('/').slice(1).includes('ostree_tree');
+      });
+      if (!hasOstreeTree && !hardwareId) {
+        alert('The selected directory has no "ostree_tree" subdirectory. Please set "Hardware ID" in the advanced options.');
+        document.getElementById('upload-advanced').open = true;
+        document.getElementById('upload-hardware-id').focus();
```
vkhoroz, line 257: I'd first focus the element and then show an alert; so that a user already sees in background what to do, and is less puzzled by our message.

vkhoroz (approved):
LGTM

PR #211, opened 2026-07-16
feat(perf-test): self-contained mTLS Locust harness + real per-device seed certs

StealthyCoder:
## What

  Adds `contrib/perf-test/`, a self-contained Locust-based mTLS performance
  test for `dg-sat`: a fast in-process Go cert generator (replacing
  `contrib/scale-test`'s openssl-fork-per-device approach), a device actor
  (mTLS `:8443`, `GET /device`), and an admin actor (Bearer token `:8080`,
  paginated `GET /v1/devices`). This is deliberately scoped down — `/config`,
  `/events`, the update-check/download flow, and task tagging/named Makefile
  targets are follow-up work in a stacked PR (see below), so this stays
  reviewable as a first increment.

  Also: `contrib/scale-test` (the superseded Python/openssl harness) is
  removed, and `cmd/seed` is updated to generate real per-device certs instead
  of a shared dummy pubkey, per review feedback that the two shouldn't
  duplicate cert-generation logic.

  ## Why a new folder instead of extending `scale-test`

  `contrib/scale-test`'s cert generation forks `openssl` per device — at scale
  that's thousands of process spawns, minutes just to get to a runnable state.
  `contrib/perf-test/gen-certs` is a single Go binary doing all keygen+signing
  in-process: 5000 ECDSA P-256 device certs in ~250ms.

  ## What it covers

  | Actor | Endpoints |
  |---|---|
  | Device (mTLS `:8443`) | `GET /device` |
  | Admin (REST `:8080`) | `GET /v1/devices` (paginated) |

  ## cmd/seed now shares cert generation with gen-certs

  `cmd/seed`'s `seedDevices()` used to register every seeded device with one
  hardcoded dummy RSA pubkey — fine for populating the UI, but nothing that
  could ever mTLS-authenticate. This PR extracts `gen-certs`'s CA/leaf
  `gen-certs` and `cmd/seed` call it:

  - `cmd/seed` now issues each device a real ECDSA leaf cert from a CA
    generated lazily on first use, appended to `cas.pem` (not overwritten, so
    `contrib/run-local.sh`'s own dev TLS leaf-as-CA keeps working alongside
    it). Cert+key land under `datadir/certs/seed-devices/<uuid>/client.pem`.
  - Device UUIDs stay on the existing deterministic `seed-device-%05d` scheme
    — only the pubkey/cert material changed, not device identity.
  - Re-running `seedDevices` against an existing datadir skips cert
    (re)generation for devices that already have one on disk, so it doesn't
    silently invalidate a cert still valid in the DB. Covered by
    `TestSeedDevicesIdempotent`.
    
  ## Design notes

  - **Admin traffic uses absolute URLs, not `self.host`.** Locust's runner
    overwrites every registered `User` class's `host` with the mTLS device
    host immediately before each spawn — `AdminUser` builds full URLs
    instead, including when following the paginated `Link` header (the
    server returns it as a relative path).
  - **`storage/devcerts` is crypto-only, no file I/O** — callers keep full
    control over on-disk layout and concurrency (`gen-certs`'s worker pool
    over `runtime.NumCPU()` is unchanged).
    
  ## Testing
  
  - `go build ./...`, `go test ./...`, `go vet ./...`, `gofmt -l .` all clean.
  - `storage/devcerts` has its own tests: CA generation, leaf issuance,
    signature verification against the issuing CA.
  - `cmd/seed`'s `TestSeed` extended to assert seeded devices get distinct
    real pubkeys (not a shared dummy), a real cert file is written, and the
    cert chains to the CA appended in `cas.pem`. New `TestSeedDevicesIdempotent`
    covers re-running against an existing datadir.
  - Manually ran `./contrib/run-local.sh` end-to-end: confirmed `cas.pem`
    contains both the original dev TLS leaf and the new seed CA, and a seeded
    device's cert verifies against it via `openssl verify`.
  - `contrib/perf-test`: built `gen-certs`, ran it standalone (device certs
    verify against `cas.pem`), and confirmed `docker compose config` parses.
   
  ## Follow-up

  A stacked PR (targeting this branch) adds `/config`, `/events`, the
  check-for-update/download flow, per-task tagging, and Makefile/profile/scene
  ergonomics. Kept separate per review feedback that the original combined PR
  was too large to review as one unit.


contrib/README.md
```diff
@@ -2,6 +2,12 @@
 
 This directory contains tools useful for local development
 
+## perf-test
+
+Self-contained Locust-based mTLS performance test that seeds and drives
+thousands of fake devices against the update server. See
+`perf-test/README.md`.
+
```
doanac, line 10: This commit comes before any actual work and the commit message is referring to my old branch that was never merged into this project

contrib/perf-test/docker-compose.yml
```diff
@@ -0,0 +1,67 @@
...
+    volumes:
+      - ${DATA_DIR:-./perf-test-data}:/data
+    command:
+      - /perf/setup.sh
+      - --datadir
+      - /data
+      - --num-devices
+      - "${NUM_DEVICES:-5000}"
+      - --hostname
+      - dg-sat
```
doanac: There are a lot of `dg-sat` references in this commit including the binary being built. The binary is now "fioserver".

contrib/perf-test/Makefile
```diff
@@ -0,0 +1,31 @@
...
+NUM_DEVICES ?= 5000
+DATA_DIR    ?= ./perf-test-data
+
+.PHONY: run setup headless clean
+
+run: ## Web UI at http://localhost:8089
+	NUM_DEVICES=$(NUM_DEVICES) DATA_DIR=$(DATA_DIR) docker compose up --build
+
+setup: ## Generate certs + seed auth only (no load test)
+	NUM_DEVICES=$(NUM_DEVICES) DATA_DIR=$(DATA_DIR) docker compose up --build setup
```
doanac, line 13: Does "run" depend on "setup"? Might put "setup" before run to make that more obvious

cmd/seed/main.go
```diff
@@ -50,11 +43,55 @@ func openStorage(datadir string) (*storage.FsHandle, *api.Storage, *gateway.Stor
 	return fs, ap, gw, nil
 }
 
+// seedCA returns a CA usable to issue seeded devices' leaf certs, appending
```
doanac: I think I'd drop this commit and the refactor one before. I thought this would reduce code/complexity. Seems like its adding complexity

doanac (commented):
None of your commits have signed-off-by or co-authored-by.

doanac (approved):

PR #212, opened 2026-07-16
feat(perf-test): update-flow, task tagging, and scenario tooling

StealthyCoder:
## What

Builds out the rest of the perf-test harness on top of #211 (merged): the
check-for-update/download flow, full steady-state device traffic, per-task
tagging for isolated scenario runs, and Makefile ergonomics around all of it.

- Full steady-state device traffic: `GET /config`, `POST /events` alongside
  the existing `GET /device` (weighted 5/2/3).
- **Check for update + download** (`UpdateFlow`): an ordered
  `SequentialTaskSet` — `timestamp.json` → `snapshot.json` → `targets.json`
  → resolve target/hash → `POST /ostree/download-urls` → `GET /ostree/config`.
- **`--seed-update`** in `gen-certs`: seeds an unsigned-but-structurally-valid
  TUF target + a real 256KiB ostree object, assigned to every generated
  device via a rollout — all before `fioserver serve` starts, so there's no
  server restart or async wait.
- **Per-task tagging** (`device:*`/`admin:*`/`update:*`/`update:check`/
  `update:download`) so `--tags`/`--exclude-tags` can isolate one scenario
  at a time instead of steady-state/admin/update traffic all competing for
  the same SQLite writer in one run.
- Named `locust-*` targets (`locust-update-check`, `locust-update`,
  `locust-admin`, `locust-steady-state`) plus a `profiles/`/`scenes/`
  composition mechanism (`make headless-scenario PROFILE=smoke
  SCENE=update-check`) so common cases don't require remembering the
  `--tags` incantation.
- **`make help`** (and bare `make`) lists every target + available
  profiles/scenes, generated from the Makefile's own doc-comments so it
  can't drift from what's implemented.
- **`DRY_RUN=1`** prints the resolved `docker compose` command instead of
  running it.

## Design notes

- **Rollouts are assigned by UUID, not group.** A device that self-registers
  via real mTLS always has `group_name=""` — a rollout keyed by `groups`
  would silently match nothing.
- **Seeded devices get their real extracted pubkey, not a placeholder** —
  a fake key would make the real mTLS handshake hit "Key rotation is not
  supported" on first contact, since the DB pubkey wouldn't match the cert
  already on disk for that device.
- **Ordered, state-dependent flows use a `SequentialTaskSet`**, state
  stashed on the TaskSet instance (not the User or a module global) so it
  doesn't leak across unrelated tasks or other simulated devices.
- **`--tags`/`--exclude-tags` alone can spawn a `User` class with zero
  tasks.** Locust prunes `.tasks` by tag but doesn't drop the `User` class
  itself from the spawn pool — always pass the relevant class name
  positionally alongside `--tags`/`--exclude-tags` (documented in README).

## Testing

- `go build ./...`, `go test ./...`, `go vet ./...`, `gofmt -l .` clean.
- `gen-certs --seed-update` run standalone: confirmed the seeded target/
  rollout is created and the device cert still verifies against `cas.pem`.
- `make help` confirmed to list all named targets and discovered
  profiles/scenes.
- Full 5000-device runs (unseeded baseline, and seeded + update-flow +
  admin mixed run): 0 failures, 0 exceptions, full device registration
  confirmed via the `Link` header's `last` offset. (Baseline/ratescan
  result docs intentionally excluded from this PR's diff, per request —
  results to follow separately.)


contrib/perf-test/Makefile
```diff
@@ -1,33 +1,156 @@
...
+	$(SETUP_CMD)
+endif
+
+up: ## Start the persistent stack (setup once, then fioserver) so headless/locust-* can run repeatedly without redoing setup. Add DRY_RUN=1 to print the command instead of running it.
+ifdef DRY_RUN
+	@echo $(UP_CMD)
+else
+	$(ensure_data_dir)
+	$(UP_CMD)
+endif
```
doanac: not sure we need the complexity of dryrun.

contrib/perf-test/README.md
```diff
@@ -1,44 +1,217 @@
...
+
+## Getting started
+
+**Web UI, one-shot:**
+
+```
+make run
+```
+
+Open <http://localhost:8089>, set the number of users and spawn rate, click
```
doanac: I think you need to explain what number of users and spawn rate does. I think number of users is the number of devices? Also - you really need SEED_UPDATE=1 or the results include failures. Maybe SEED_UPDATE should always be set?

contrib/perf-test/README.md
```diff
@@ -1,44 +1,217 @@
...
+  `update:download` (the last two only). See "Running isolated scenarios"
+  below — this flow only succeeds once a rollout has been seeded (see
+  "Seeding a TUF target"), so a default/unseeded run should pass
+  `--exclude-tags update` or it will 404/400 on every step.
+
+## Running isolated scenarios
 
-## Quick start
+Every task is tagged (mirroring the `@locust.tag(...)` convention in
+`test-dspl`'s `hub_registry.py`), so you can run one scenario at a time —
```
doanac: test-dspl shouldn't be mentioned

contrib/perf-test/README.md
```diff
@@ -1,44 +1,217 @@
...
+`profiles/`) are named scale/timing presets (`NUM_DEVICES`/`SPAWN_RATE`/
+`RUN_TIME`); **scenes** (under `scenes/`) are named task-selection presets
+(`LOCUST_ARGS`, i.e. the same `--tags`/`User`-class combinations as the
+named targets above). Compose either or both via `PROFILE=`/`SCENE=` on the
+`headless-scenario` target:
+
+```
+make headless-scenario PROFILE=smoke SCENE=update-check
+make headless-scenario PROFILE=full  SCENE=admin-only SEED_UPDATE=1
 ```
```
doanac, line 278: At this point we have documented scenarios, targets, scenes. Its all necessary, but maybe there needs to be a section up higher to the effect of "common tasks". Then cover some stuff that's useful. Strawman tasks:
* 5000 devices - with admin listing and devices checking-in
* 5000 devices doing updates
* Some tests to target a specific endpoint to find out its max requests per second?

StealthyCoder: Added these and extended some more scenarios further in the documentation.

doanac (approved):
LGTM - there's nothing we can't come back and fix later.

PR #213, opened 2026-07-17
Add support for lmp-device-register

doanac:
This set of changes adds support for the LmP's lmp-device-register if the user provides the "-T <token>" option.

Support for OAuth2 Device Flow will follow in a separate PR.

Fixes issue #207

### DIFF
--- a/auth/provider_common.go
+++ b/auth/provider_common.go
@@ -38,15 +38,22 @@ func (p *commonProvider) DropSession(c echo.Context, session *Session) {
 
 func (p *commonProvider) GetUser(c echo.Context) (*users.User, error) {
 	authHeader := c.Request().Header.Get("Authorization")
-	if len(authHeader) > 0 {
+	authToken := ""
+	if len(authHeader) == 0 {
+		authToken = c.Request().Header.Get("OSF-TOKEN")
+	}
+	if len(authHeader) > 0 || len(authToken) > 0 {
 		if err := p.rateLimiter.allow(c.RealIP()); err != nil {
 			return nil, server.EchoError(c, err, http.StatusTooManyRequests, err.Error())
 		}
-		parts := strings.SplitN(authHeader, " ", 2)
-		if len(parts) != 2 || strings.ToLower(parts[0]) != "bearer" {
-			return nil, fmt.Errorf("invalid authorization header")
+		if len(authHeader) > 0 {
+			parts := strings.SplitN(authHeader, " ", 2)
+			if len(parts) != 2 || strings.ToLower(parts[0]) != "bearer" {
+				return nil, fmt.Errorf("invalid authorization header")
+			}
+			authToken = parts[1]
 		}
-		user, err := p.users.GetByToken(parts[1])
+		user, err := p.users.GetByToken(authToken)
 		if err != nil {
 			p.rateLimiter.FlagBadOperation(c)
 			slog.Warn("unable to get user by token", "error", err)
--- a/server/ui/api/handlers.go
+++ b/server/ui/api/handlers.go
@@ -6,6 +6,7 @@ package api
 import (
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
+	"golang.org/x/time/rate"
 
 	"github.com/foundriesio/update-server/auth"
 	"github.com/foundriesio/update-server/server"
@@ -14,13 +15,18 @@ import (
 )
 
 type handlers struct {
+	ca      *DeviceCa
 	storage *storage.Storage
 }
 
 var EchoError = server.EchoError
 
-func RegisterHandlers(e *echo.Echo, storage *storage.Storage, a auth.Provider) {
-	h := handlers{storage: storage}
+func RegisterHandlers(e *echo.Echo, ca *DeviceCa, storage *storage.Storage, a auth.Provider) {
+	h := handlers{
+		ca:      ca,
+		storage: storage,
+	}
+
 	g := e.Group("/v1")
 	g.Use(authUser(a))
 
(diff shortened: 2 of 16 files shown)


server/ui/api/handlers.go
```diff
@@ -37,6 +45,12 @@ func RegisterHandlers(e *echo.Echo, storage *storage.Storage, a auth.Provider) {
 	g.GET("/configs/device/:uuid/applied", h.deviceAppliedConfigsGet, requireScope(users.ScopeDevicesR))
 	g.GET("/configs/device/:uuid/history", h.deviceConfigsHistory, requireScope(users.ScopeDevicesR))
 	g.GET("/devices", h.deviceList, requireScope(users.ScopeDevicesR))
+	if ca != nil {
+		slog.Info("Device CA is configured, enabling device registration endpoint")
```
vkhoroz: I'm starting to see a bad habit of using a global `slog` logger sporadically, when we should be using our context logger instead (rich context and uniform format).

I see 3 ways to fix this:
1. Pass context into this method and get logger from it using `CtxGetLog`.
2. Use `e.StdLogger` (which uses an adapter to log via our context logger).
3. (Probably the best) print a simple log message in `server/ui/server.go`, where you load that CA and context is accessible.

(I found the other instance where `slog` is used here: https://github.com/foundriesio/update-server/blob/main/server/ui/server.go#L51; but we have our context accessible, so should use it.)

server/ui/server.go
```diff
@@ -52,9 +52,14 @@ func NewServer(ctx context.Context, db *storage.DbHandle, fs *storage.FsHandle,
 
 	daemons := daemons.New(ctx, strg, users)
 
+	ca, err := apiHandlers.LoadDeviceCa(fs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load device CA: %w", err)
+	}
```
vkhoroz, line 60: I meant to do `CtxGetLog(ctx).Log("Device CA is configured")` - that's more than enough for the debug message.

server/ui/api/handlers_devices_create.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	first, rest := pem.Decode(buf)
+	if first == nil || len(strings.TrimSpace(string(rest))) > 0 {
+		return nil, fmt.Errorf("malformed PEM data for %s", storage.CertsDeviceCaKeyFile)
+	}
```
vkhoroz: One more instance of that pem parsing.
Copying itself wouldn't be bad; but I see that some copy instances forget to trim, error format is fluent, etc.

server/ui/api/handlers_devices_create.go
```diff
@@ -0,0 +1,136 @@
...
+	rootCrt := string(buf)
+
+	// load dg tls cert to get the DG URL
+	buf, err = fs.Certs.ReadFile(storage.CertsTlsPemFile)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil, nil
+		}
+		return nil, err
```
vkhoroz, line 71: I think we ca do better:
- If device CA is absent - device registration should be disabled.
- If device CA is present but other certs are absent - that's an error.
- If all certs are present - device registration should be enabled.

server/ui/api/handlers_devices_create.go
```diff
@@ -98,6 +102,33 @@ func LoadDeviceCa(fs *storage.FsHandle) (*DeviceCa, error) {
...
+	}
+	serial, err := rand.Int(rand.Reader, new(big.Int).Lsh(big.NewInt(1), 128))
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate serial number: %w", err)
+	}
+	template := x509.Certificate{
+		SerialNumber:          serial,
+		Subject:               csr.Subject,
+		NotBefore:             time.Now(),
+		NotAfter:              time.Now().Add(7300 * 24 * time.Hour),
```
vkhoroz, line 98: Do we really want client certs to expire in 20 years?
Maybe, we should limit them by the Device CA expiration?
Maybe, make that configurable and do `min(configured_value, deviceCA.NotAfrer)`?

storage/api/api_storage.go
```diff
@@ -804,6 +810,24 @@ func (s *stmtDeviceDelete) run(uuid string) error {
 	return err
 }
 
+type stmtDevicePut storage.DbStmt
+
+func (s *stmtDevicePut) Init(db storage.DbHandle) (err error) {
+	s.Stmt, err = db.Prepare("apiDevicePut", `
+		INSERT INTO devices (uuid, pubkey, labels, deleted, created_at)
+		VALUES (?, ?, ?, 0, unixepoch())`)
```
vkhoroz: Can you set use `false` for the `deleted` column, and also explicitly set `last_seen` to 0?

auth/provider_common.go
```diff
@@ -38,15 +38,22 @@ func (p *commonProvider) DropSession(c echo.Context, session *Session) {
 
 func (p *commonProvider) GetUser(c echo.Context) (*users.User, error) {
 	authHeader := c.Request().Header.Get("Authorization")
-	if len(authHeader) > 0 {
+	authToken := ""
+	if len(authHeader) == 0 {
+		authToken = c.Request().Header.Get("OSF-TOKEN")
+	}
```
vkhoroz, line 44: Is it safe to support OSF-TOKEN for all endpoints?
I think it is; just wouldn't want us to fall into some tar pit.

vkhoroz (commented):
Looks good. Just some minor findings.

server/ui/api/handlers_devices_create.go
doanac: We've concluded that if you are sophisticated enough to want more conservative expiry dates, then you are sophisticated enough to either fork or build your provisioning system.

auth/provider_common.go
doanac: I didn't like doing it, but it doesn't make things less safe.

storage/api/api_storage.go
doanac: fixed in force

server/ui/api/handlers_devices_create.go
doanac: fixed in force

server/ui/server.go
doanac: fixed in force

server/ui/api/handlers_devices_create.go
doanac: I added commit [40df471](https://github.com/foundriesio/update-server/pull/213/commits/40df47166ea3eeba3ba2cafcf76077d6bda5371e)

server/ui/server.go
```diff
@@ -48,13 +47,20 @@ func NewServer(ctx context.Context, db *storage.DbHandle, fs *storage.FsHandle,
 	if err != nil {
 		return nil, err
 	}
-	slog.Info("Using authentication provider", "name", provider.Name())
+	context.CtxGetLog(ctx).Info("Using authentication provider", "name", provider.Name())
```
vkhoroz: Usually, we do `log := context.CtxGetLog(ctx)` at the function start if there is more than one log line; like here.

vkhoroz (approved):
LGTM

server/ui/server.go
doanac: fixed in force push

PR #214, opened 2026-07-17
cli: Include TUF details in `updates show`

doanac:
This makes the CLI show more helpful information like our web UI was doing.

### DIFF
--- a/cli/subcommands/updates/show.go
+++ b/cli/subcommands/updates/show.go
@@ -4,17 +4,20 @@
 package updates
 
 import (
+	"encoding/json"
 	"fmt"
+	"sort"
 
 	"github.com/foundriesio/update-server/cli/api"
-	"github.com/foundriesio/update-server/cli/subcommands"
 	"github.com/spf13/cobra"
 )
 
+var showTufDetails bool
+
 var showCmd = &cobra.Command{
 	Use:   "show <tag> <update-name>",
-	Short: "Show rollouts for an update",
-	Long:  `Display all rollouts for a specific update`,
+	Short: "Show details for an update",
+	Long:  `Display details about an update from its TUF metadata`,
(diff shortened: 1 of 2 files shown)


cli/subcommands/updates/show.go
```diff
@@ -24,26 +27,173 @@ var showCmd = &cobra.Command{
...
+}
+
+func printApps(custom map[string]any) {
+	apps := make(map[string]string)
+	if custom != nil {
+		if dockerApps, ok := custom["docker_compose_apps"].(map[string]any); ok {
+			for name, val := range dockerApps {
+				if appMap, ok := val.(map[string]any); ok {
+					if uri, ok := appMap["uri"].(string); ok {
+						apps[name] = uri
```
vkhoroz, line 181: This and other methods are rather painful when it comes to parsing.

Maybe, there is a way to leverage the JSON parser for this. For example, this recipe should work:
- Define a `TufTargetCustomData` struct with only fields we care about;
- Leverage a fact that `Custom` [attribute is a json.RawMessage](https://github.com/foundriesio/update-server/blob/main/storage/tuf/tuf_data.go#L141):
  * for "raw" print purposes, we just print it as a JSON.
  * when we need to extract fields, we unmarshal it into a `TufTargetCustomData`, and operate individual fields.

vkhoroz (approved):
LGTM

I don't like unmarshalling custom JSON into a map, and casting each individual field.
I think we can do better with a regular json parsing into a struct.
I'll leave it up to you to decide, as this feature is good with a bit quirky code.

cli/subcommands/updates/show.go
doanac: A year ago, i would spend time on doing that. With Claude doing work on "leaf" code like this, I've stopped bothering or caring.

PR #215, opened 2026-07-20
Add basic e2e testing

doanac:
This creates a simple way to do some primitive e2e testing of the update server using a "device" running fioup inside a docker:dind container. 

Once we agree on the general functionality of this, I'll propose some sort of periodic cron-based task to run these tests.

### DIFF
--- a/contrib/e2e/conftest.py
+++ b/contrib/e2e/conftest.py
@@ -254,23 +254,6 @@ def registered_device(update_server, fioup_device) -> dict:
 def update_server(request, fioserver_bin):
     """Generate PKI, start update-server; yield datadir Path."""
     datadir = Path(tempfile.mkdtemp(prefix="fioserver-"))
-    gen_pki = REPO_ROOT / "gen_pki.sh"
-
-    print("\n[setup] Generating PKI ...", flush=True)
-    result = subprocess.run(
-        [
-            "bash",
-            str(gen_pki),
-            str(datadir),
-            str(fioserver_bin),
-            "update-server",
-            "e2e-factory",
-        ],
-        check=True,
-        capture_output=True,
-        text=True,
-    )
-    print(result.stdout)
 
     print("[setup] Initialising auth (noauth/test mode) ...", flush=True)
     subprocess.run(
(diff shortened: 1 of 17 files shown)


contrib/e2e/test_remote_actions.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+"""E2E test: remote actions — run a command on a device, verify via CLI and web UI."""
+
+SERVER_URL = "http://localhost:8080"
```
StealthyCoder, line 6: Probably does not matter, but we call update-server in the `test_connection.py`, should we also call it here?

contrib/e2e/test_connection.py
```diff
@@ -0,0 +1,25 @@
...
+    last_seen = registered_device.get("last-seen", 0)
+    assert (
+        last_seen > 0
+    ), f"Device registered but last-seen is zero: {registered_device}"
+
+
+def test_device_register(fioup_device, fiocli):
+    uuid = "11111111-1111-1111-1111-111111111111"
+    fioup_device.run("rm -rf /var/sota/*")
+    fioup_device.run(f"DEVICE_API=http://update-server:8080/v1/devices fioup --verbose register --apps ' ' --uuid={uuid} --factory=e2e-factory --name=e2e-test-device --tag=main --api-token doesnotmatter")
```
StealthyCoder, line 21: Should this reference then also be to `fioserver` or is it okay to use this name for the container itself?

contrib/e2e/conftest.py
StealthyCoder, line 1: I would prefer to use the `docker` package rather than the `subprocess` module. In `fioup_device` there are three different ways/styles of invoking `docker`, that might give more noise in the future. With `docker` package API is consistent the same everywhere. That is just my view on it.

contrib/e2e/test_remote_actions.py
doanac: I might be misunderstanding you but: its "localhost" for the user facing API. Its "update-server" for the fioup device

contrib/e2e/test_connection.py
doanac: in this case (from inside the container) it knows the API by this hostname.

contrib/e2e/conftest.py
doanac: this was a great idea. It also runs things a little faster. I've force pushed the change - its basically redoing a bunch of the first two commits in the series

contrib/e2e/test_remote_actions.py
StealthyCoder: Depends on where you run, but I understand your distinction now. Consider this resolved. :+1:

StealthyCoder (approved):
This looks great. Only thing are the GitHub security flagged ones if you agree with their assessment. However for this setup it looks very nice.

PR #216, opened 2026-07-21
Add support for deleting old Updates

doanac:
@vkhoroz - the last two commits were suggested by Claude code review. The one for atomic/race-free deletes makes things a little more complex. Let me know if you want those included in this patch set or not.

Fixes issue #208

### DIFF
--- a/storage/api/api_storage.go
+++ b/storage/api/api_storage.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"iter"
 	"log/slog"
+	"os"
 	"slices"
 	"strings"
 
@@ -72,6 +73,7 @@ var (
 	IsDbError             = storage.IsDbError
 	ErrDbConstraintUnique = storage.ErrDbConstraintUnique
 	ErrInvalidUpdate      = storage.ErrInvalidUpdate
+	ErrUpdateInUse        = storage.ErrUpdateInUse
 )
 
 // DeviceListOpts lets you set the order devices will be returned
@@ -130,6 +132,7 @@ type Storage struct {
 	stmtUndenyDevice     stmtUndenyDevice
 	stmtDeviceSetLabels  stmtDeviceSetLabels
 	stmtDeviceSetUpdate  stmtDeviceSetUpdate
+	stmtUpdateDelete     stmtUpdateDelete
 	stmtUpdateInsert     stmtUpdateInsert
 	stmtUpdateList       stmtUpdateList
 }
@@ -214,6 +217,7 @@ func NewStorage(db *storage.DbHandle, fs *storage.FsHandle) (*Storage, error) {
 		&handle.stmtDevicePut,
 		&handle.stmtUpdateInsert,
 		&handle.stmtUpdateList,
+		&handle.stmtUpdateDelete,
 	); err != nil {
 		return nil, err
 	}
@@ -356,6 +360,20 @@ func (s Storage) ListUpdates(tag string) (map[string][]Update, error) {
 
 func (s Storage) DeviceCreate(uuid, pubkey string, labels Labels) error {
 	return s.stmtDevicePut.run(uuid, pubkey, labels)
+}
+
+// DeleteUpdate removes an update's database row and on-disk directory. It
+// refuses (returning ErrUpdateInUse) if any non-denied device is still assigned
+// to the update. It returns ErrNotExist if the update does not exist.
+func (s Storage) DeleteUpdate(tag, name string) error {
+	existed, err := s.stmtUpdateDelete.run(tag, name)
+	if err != nil {
+		return err
+	}
+	if !existed {
+		return os.ErrNotExist
+	}
+	return s.fs.Updates.Delete(tag, name)
 }
 
 func (s Storage) GetUpdateTufMetadata(tag, updateName string) (map[string]map[string]any, error) {
@@ -877,4 +895,29 @@ func (s *stmtUpdateList) run(tag string) (map[string][]Update, error) {
 		res[t] = append(res[t], u)
 	}
 	return res, rows.Err()
+}
+
+type stmtUpdateDelete storage.DbStmt
+
+func (s *stmtUpdateDelete) Init(db storage.DbHandle) (err error) {
+	s.Stmt, err = db.Prepare("apiUpdateDelete", `
+		DELETE FROM updates 
+		WHERE tag = ?
+  		AND name = ?;`)
+	return
+}
+
+func (s *stmtUpdateDelete) run(tag, name string) (bool, error) {
+	res, err := s.Stmt.Exec(tag, name)
+	if err != nil {
+		if err.Error() == ErrUpdateInUse.Error() {
+			return false, ErrUpdateInUse
+		}
+		return false, err
+	}
+	rowsAffected, err := res.RowsAffected()
+	if err != nil {
+		return false, err
+	}
+	return rowsAffected > 0, nil
 }
--- a/storage/api/api_storage.go
+++ b/storage/api/api_storage.go
@@ -851,9 +851,13 @@ type stmtUpdateList storage.DbStmt
 
 func (s *stmtUpdateList) Init(db storage.DbHandle) (err error) {
 	s.Stmt, err = db.Prepare("apiUpdateList", `
-		SELECT tag, name, uploaded_at, uploaded_by FROM updates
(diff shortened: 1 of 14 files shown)


storage/api/api_storage.go
```diff
@@ -851,9 +851,13 @@ type stmtUpdateList storage.DbStmt
...
 	s.Stmt, err = db.Prepare("apiUpdateList", `
-		SELECT tag, name, uploaded_at, uploaded_by FROM updates
-		WHERE (? = '' OR tag = ?)
-		ORDER BY tag, uploaded_at, name`)
+		SELECT u.tag, u.name, u.uploaded_at, u.uploaded_by, COUNT(d.uuid)
+		FROM updates u
+		LEFT JOIN devices d
+			ON d.tag = u.tag AND d.update_name = u.name AND d.deleted = false
+		WHERE (? = '' OR u.tag = ?)
+		GROUP BY u.tag, u.name, u.uploaded_at, u.uploaded_by
```
vkhoroz: As (tag, name) is a primary index on updates, only these two fields are needed in the group by clause.
SQL standard allows for dependency injection for that specific use case, and sqlite planner should support it.

storage/api/api_storage.go
```diff
@@ -358,6 +362,38 @@ func (s Storage) DeviceCreate(uuid, pubkey string, labels Labels) error {
...
+						return ErrUpdateInUse
+					}
+					break
+				}
+			}
+
+		} else {
+			return os.ErrNotExist
+		}
+	}
```
vkhoroz: This is strange.
A simple `GetUpdate(tag, name)` would be a primary index lookup, even if we include the `device_count` aggregate into it.

Hint: you may have an internal `listUpdates(tag, name)` which does the heavy lifting, with `ListUpdates(tag)` and `GetUpdates(tag, name)` both calling it; if you assume that empty string for a name means all names (just like we do for a tag). If `name` is non-empty - `listUpdates` would return either zero or one item for a corresponding tag.

cmd/server/main.go
```diff
@@ -18,14 +18,15 @@ type VersionCmd struct{}
...
-	TufInit  *TufInitCmd  `arg:"subcommand:tuf-init" help:"Initialize TUF keys and root metadata for this server"`
-	UserAdd  *UserAddCmd  `arg:"subcommand:user-add" help:"Add a new user if local authentication is enabled"`
-	Version  *VersionCmd  `arg:"subcommand:version" help:"Print the version of the program"`
+	AuthInit  *AuthInitCmd  `arg:"subcommand:auth-init" help:"Initialize authentication configuration for this server"`
+	Csr       *CsrCmd       `arg:"subcommand:create-csr" help:"Create a TLS certificate signing request for this server"`
+	SignCsr   *CsrSignCmd   `arg:"subcommand:sign-csr" help:"Create the TLS certificate from the signing request"`
+	PkiInit   *PkiInitCmd   `arg:"subcommand:pki-init" help:"Bootstrap a complete PKI (root CA, TLS cert, and device CA) from scratch"`
+	Serve     *ServeCmd     `arg:"subcommand:serve" help:"Run the REST API and device-gateway services"`
+	TufInit   *TufInitCmd   `arg:"subcommand:tuf-init" help:"Initialize TUF keys and root metadata for this server"`
+	UpdatesGc *UpdatesGcCmd `arg:"subcommand:updates-gc" help:"Delete updates that have no assigned devices"`
```
vkhoroz: I'd rather do this as a daemon running in background and firing e.g. daily.
We do already have a machinery for the API daemons; so it should be easy.

cmd/server/updates_gc.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	tags, err := apiStorage.ListUpdates("")
+	if err != nil {
+		return fmt.Errorf("failed to list updates: %w", err)
+	}
+
+	for tag, updates := range tags {
+		for _, u := range updates {
+			if u.DeviceCount > 0 {
```
vkhoroz: We need to have a cutoff based on the update time; so that we don't end up deleting fresh updates accidentally. I'd make that cutoff fairly long, e.g. a week or even a month.

storage/api/api_storage.go
```diff
@@ -365,32 +363,43 @@ func (s Storage) DeviceCreate(uuid, pubkey string, labels Labels) error {
...
+		).Scan(&deviceCount); err != nil {
+			return err
+		}
+		if deviceCount > 0 {
+			return ErrUpdateInUse
+		}
+
+		_, err = tx.Exec(`DELETE FROM updates WHERE tag=? AND name=?`, tag, name)
+		return err
+	}); err != nil {
```
vkhoroz: The code needed to make this transaction is not a problem.
A problem is that it will block the database for the entire transaction duration (blocking any other writes).

Another problem is a maintenance - now we need to always keep this transaction in mind when adding new features.
And benefit from it is purely theoretic (to make some random AI happy).

cmd/server/updates_gc.go
```diff
@@ -30,16 +34,29 @@ func (c UpdatesGcCmd) Run(args CommonArgs) error {
 		return fmt.Errorf("failed to initialize storage: %w", err)
 	}
 
+	excluded := make(map[string]bool, len(c.Excludes))
+	for _, e := range c.Excludes {
+		excluded[e] = true
+	}
```
vkhoroz: I think you've nailed the same problem I described earlier (this garbage collector thrashes too many false positives).
But, instead of a cure, this commit proposes a bandage.
I doubt a regular user would ever assume any problem with this garbage collector before it is too late.

vkhoroz (commented):
Overall looks good.

I'd throw that transaction (no value), and change how we garbage collect.

cmd/server/updates_gc.go
doanac: I'm starting to lean towards dropping this GC logic for now until we start to use the product more and understand it better. Additionally - you wanted it to auto gc whereas I was scared to do that. I think maybe I'll just dropped the last few commits in this series and just focus on adding the ability to delete an update.

vkhoroz: Yes, that is probably the best option for now.

storage/api/api_storage.go
doanac: updated in force push. take a look and see if you agree.

storage/api/api_storage.go
doanac: I wound up going a different direction on this and let sqlite do the heavy lifting using a trigger. Take a look at commit f7b99ac5b1684b3402ad29c1307e6570d53a8f83 to see what you think

storage/users/scopes.go
```diff
@@ -31,6 +31,7 @@ const (
```
vkhoroz, line 31: Typo in commit message: `storge->storage`.

storage/db.go
```diff
@@ -165,6 +165,20 @@ func createTables(db *sql.DB) error {
...
+			FOR EACH ROW
+			WHEN EXISTS (
+				SELECT 1
+				FROM devices
+				WHERE tag = OLD.tag
+				AND update_name = OLD.name
+				AND deleted = 0
+			)
+			BEGIN
+				SELECT RAISE(ABORT, 'update is referenced by one or more devices');
```
vkhoroz: This deserves a constant, as it is used in 3 places.

I'd define that `ErrUpdateInUse = errors.New("update is referenced by one or more devices")` in this module, and use `fmt.Sprintf` (or string concatenation) to inline it into the SQL statement.
Then, on the API level you only need an alias, and compare a string as `err.Error() == ErrUpdateInUse.Error()`.

storage/db.go
```diff
@@ -165,6 +165,20 @@ func createTables(db *sql.DB) error {
...
+		CREATE TRIGGER prevent_delete_update_in_use
+			BEFORE DELETE ON updates
+			FOR EACH ROW
+			WHEN EXISTS (
+				SELECT 1
+				FROM devices
+				WHERE tag = OLD.tag
+				AND update_name = OLD.name
+				AND deleted = 0
+			)
```
vkhoroz, line 179: This looks pretty clever.
I only wonder if we'd ever want to do the "force" feature for this (I bet we don't).

vkhoroz (approved):
LGTM

storage/db.go
doanac: good idea. Fixed in force push

PR #222, opened 2026-07-28
docs: Introduce section about TUF

doanac:


docs/tuf.md
```diff
@@ -0,0 +1,64 @@
...
+> [!NOTE]
+> The [Quick Start](./quick-start.md) shares how to initialize TUF keys
+> and metadata.
+
+## Root role
+
+The update server uses a single Root key under `<datadir>/tuf/keys/root.key`.
+It is encrypted on disk with an HMAC key. This key **cannot be lost**.
+Loss of this key and/or the HMAC key will make it impossible to sign new
+Updates after the key **expires**.
```
vkhoroz:
```suggestion
The update server uses a single Root key under `<datadir>/tuf/keys/root.key`.
It is encrypted on disk with an HMAC key.
You **must never lose** this key.
A loss of a Root key and/or the HMAC key would make it impossible to add new keys;
in which case there will be no way to sign new Updates after a Root role expires.
```

Corrections rationale:
- A **key cannot be lost** may look like it is resilient to loss. I guess a stronger indicative language is more appropriate here.
- A loss of a root key instantly makes it impossible to add any new keys; and after a Root **role** expires, new updates are also impossible.  I guess a distinction between a key and a role is important here, as a key itself does not expire.

docs/tuf.md
```diff
@@ -0,0 +1,64 @@
...
+
+The root role is valid for 20 years (root.json's `signed.expires` attribute).
```
vkhoroz, line 22: I'd place this at the top of this section.
As is, it breaks up two paragraphs talking about how important a root key is.

docs/tuf.md
```diff
@@ -0,0 +1,64 @@
...
+
+> [!NOTE]
+> Backup `<datadir>/tuf/keys/root.key` **and** `<datadir>/auth/hmac.secret` to
+> multiple places including [AWS Secret Manager](https://docs.aws.amazon.com/secretsmanager/),
+> [BitWarden](https://bitwarden.com/), and a physical copy on paper stored
+> in a safe.
```
vkhoroz: I'm not sure if I'd recommend storing a _company's root of trust_ in an online service which is only protected by a _personal_ password (or personal 2FA at most).
An AWS secret manager seems appropriate here, but probably not a BitWarden.

Also, I'd state _such as_ instead of _including_, as a user may have other preferences for cloud services.

docs/tuf.md
vkhoroz, line 22: I'm not sure if making a root role valid for 20 years was a good decision.
This way, we allow a user who stole any key to perform a freeze attack for this whole period.
Uptane recommend root role to expire in 1 year; and this is what we were doing too for FoundriesFactory.

As this is configurable, a user can set it to 20 or 100 years; but a 1 year default is reasonable enough (an industry standard).

docs/tuf.md
```diff
@@ -0,0 +1,64 @@
...
+You can remove this key from the server once it's been backed up. It is
+considered an "offline" key and is not required by the update server for
+normal operations.
+
+## Timestamp role
+
+The update server uses a single Timestamp key under `<datadir>/tuf/keys/timestamp.key`.
+It is encrypted on disk with an HMAC key. This is considered an "online" key
+and must be present for the update server to perform normal operations. Loss
+of this key is not a big deal. It is valid for 20 years.
```
vkhoroz: It is unclear why a loss of this key is not a big deal, especially given that it is required by a server for its normal operations.

Instead, I'd say something about a possibility to re-generate this key in case if it was lost.

Also, individual keys don't expire; it is the entire Root role (with all keys) that expires.

_An off-topic..._
Actually, now I think that it would be a good practice to generate a new pair of timestamp/snapshot keys every time a user rotates a root key.
We may keep two latest timestamp/snapshot key IDs in a root.json, so that existing updates remain valid.

docs/tuf.md
```diff
@@ -0,0 +1,64 @@
...
+Each Update has its own timestamp.json file. This file is signed with a
+1-week expiration value. There is a background task running every 4 hours
+that will refresh timestamps expiring in the next 24 hours.
+
+## Snapshot role
+
+The update server uses a single Snapshot key under `<datadir>/tuf/keys/snapshot.key`.
+It is encrypted on disk with an HMAC key. This is considered an "online" key
+and must be present for the update server to perform normal operations. Loss
+of this key is not a big deal. It is valid for 20 years.
```
vkhoroz: Same as above on key loss and expiration.

docs/tuf.md
```diff
@@ -0,0 +1,64 @@
...
+
+Each Update has its own snapshot.json file. This file is signed with a
+90-day expiration value that matches the Targets expiration described below.
```
vkhoroz: Targets expiration is described below, not above.
Also, I think it is more appropriate to say that _it is signed with an expiration value that matches targets expiration_ (without mentioning the 90-day period, to remove extra vagueness about what number takes precedence).

docs/tuf.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Targets role
+
+The update server uses a single Targets key under `<datadir>/tuf/keys/targets.key`.
+It is encrypted on disk with an HMAC key. This is considered an "online" key
+and must be present for the update server to perform normal operations. Loss
+of this key is not a big deal. It is valid for 20 years.
```
vkhoroz: Same as above about key loss and expiration.

docs/tuf.md
```diff
@@ -0,0 +1,64 @@
...
+
+This key is used during Update creation to sign the generated targets.json
+metadata. The metadata is signed with a 90-day expiration. See
+[Updates](./updates.md) for more details.
+
+The Snapshot and Targets expiration values are not automatically refreshed.
+It is up to the operator to keep these values fresh for Updates used over
+90 days.
```
vkhoroz, line 55: IIRC a user may override the (default) 90-day expiration period, right?

doanac: its configurable but hard-coded in a way that would require a change. I think we can do this as part of our work to add "refresh target expiration" functionality.

docs/tuf.md
doanac: I'm trying to figure out the best wording here. e.g. I don't want to put too many words in front of "you must never lose" or the user will skip right past the most important text of the section.

@kprosise - let me know what you think works best here and I'll apply it for all the sections.

docs/tuf.md
doanac: I've tried to address this with commit f2584137f3f1fa20e94666fc78bab5a384cb4bac

docs/tuf.md
doanac: I've tried to address this with commit f258413

vkhoroz (approved):
LGTM

I think this is fine to go after we work through a couple of points and Katrina gives it a read.

docs/tuf.md
```diff
@@ -0,0 +1,65 @@
...
+
+The update server leverages TUF to ensure secure OTA delivery. If you
+are unfamiliar with TUF, you can learn more from this
+[overview](https://theupdateframework.io/docs/overview/).
+
+> [!NOTE]
+> The [Quick Start](./quick-start.md) shares how to initialize TUF keys
+> and metadata.
+
+## Root role
```
kprosise:
```suggestion
## Root Role
```

docs/tuf.md
kprosise: Well for line 19, I think placing it as the last line in the subsection would be best. As for the "you must never lose" bit,  perhaps placing it in a warning callout (`> [!WARNING]`) would work? I know it precedes a note callout, but I think it would be okay to have two callouts together.

docs/tuf.md
```diff
@@ -0,0 +1,65 @@
...
+> Backup `<datadir>/tuf/keys/root.key` **and** `<datadir>/auth/hmac.secret` to
+> multiple places including [AWS Secret Manager](https://docs.aws.amazon.com/secretsmanager/),
+> [BitWarden](https://bitwarden.com/), and a physical copy on paper stored
+> in a safe.
+
+You can remove this key from the server once it's been backed up. It is
+considered an "offline" key and is not required by the update server for
+normal operations.
+
+## Timestamp role
```
kprosise:
```suggestion
## Timestamp Role
```

docs/tuf.md
```diff
@@ -0,0 +1,65 @@
...
+The update server uses a single Timestamp key under `<datadir>/tuf/keys/timestamp.key`.
+It is encrypted on disk with an HMAC key. This is considered an "online" key
+and must be present for the update server to perform normal operations. This
+key can be regenerated and is valid for 20 years.
+
+Each Update has its own timestamp.json file. This file is signed with a
+1-week expiration value. There is a background task running every 4 hours
+that will refresh timestamps expiring in the next 24 hours.
+
+## Snapshot role
```
kprosise:
```suggestion
## Snapshot Role
```

docs/tuf.md
```diff
@@ -0,0 +1,65 @@
...
+
+The update server uses a single Snapshot key under `<datadir>/tuf/keys/snapshot.key`.
+It is encrypted on disk with an HMAC key. This is considered an "online" key
+and must be present for the update server to perform normal operations. This
+key can be regenerated and is valid for 20 years.
+
+Each Update has its own snapshot.json file. This file is signed with a
+expiration value that matches the Targets expiration.
+
+## Targets role
```
kprosise:
```suggestion
## Targets Role
```

docs/tuf.md
```diff
@@ -0,0 +1,65 @@
...
+You can remove this key from the server once it's been backed up. It is
+considered an "offline" key and is not required by the update server for
+normal operations.
+
+## Timestamp role
+
+The update server uses a single Timestamp key under `<datadir>/tuf/keys/timestamp.key`.
+It is encrypted on disk with an HMAC key. This is considered an "online" key
+and must be present for the update server to perform normal operations. This
+key can be regenerated and is valid for 20 years.
```
kprosise: > "It is encrypted on disk with an HMAC key. This is considered an "online" key
and must be present for the update server to perform normal operations. This
key can be regenerated and is valid for 20 years."

Given how this bit repeats for each role, and the only thing that changes is the key name, Would it be appropriate to "factor it out" and place it at the top of the page in a generic form?

kprosise (approved):
Left some suggestions and a recommendation, but overall it looks good.

docs/tuf.md
```diff
@@ -0,0 +1,65 @@
...
+You **must never lose** this key.
+A loss of a Root key and/or the HMAC key would make it impossible to add new keys;
+in which case there will be no way to sign new Updates after a Root role expires.
+
+The root role is valid for 20 years (root.json's `signed.expires` attribute).
+
+> [!NOTE]
+> Backup `<datadir>/tuf/keys/root.key` **and** `<datadir>/auth/hmac.secret` to
+> multiple places including [AWS Secret Manager](https://docs.aws.amazon.com/secretsmanager/),
+> [BitWarden](https://bitwarden.com/), and a physical copy on paper stored
```
doanac:
```suggestion
> and a physical copy on paper stored
```

docs/tuf.md
doanac: @kprosise  - I've tried to update this with commit a5316de93cbd41d26052a70f477e7560c8b72689 - let me know if this was what you were wanting.

docs/tuf.md
doanac: I tried to clean this up with commit a5316de93cbd41d26052a70f477e7560c8b72689 - let me know what you think; i'm not super happy with my wording

kprosise (approved):
lgtm

PR #230, opened 2026-08-13
Aws enhancements

doanac:
Adds logic to enable IPV6; fixes issue #226 
Adds logic to allow streaming of fioserver logs to cloudwatch

### DIFF
--- a/contrib/terraform/aws/README.md
+++ b/contrib/terraform/aws/README.md
@@ -18,7 +18,25 @@ Two topologies are provided:
 | DNS Names | two (UI and gateway for the 2 LBs) | one |
 | Extra cost | yes | no |
 
-Both share the same modules and the same AMI.
+Both share the same modules and the same AMI. Both are dual-stack by default:
+the VPC gets an Amazon-provided IPv6 CIDR, subnets get a `/64` each, and every
+security-group rule that allows `0.0.0.0/0` gets an `::/0` sibling. Set
+`enable_ipv6 = false` to opt out and stay IPv4-only.
+
+In the load-balancer topology, the ALB and NLB run in `dualstack` mode and
+Route53 gets an `AAAA` alias record next to each `A` record; the instance
+itself stays IPv4-only, since the load balancers already terminate IPv6
+client connections and forward to its private IPv4 address — giving the
+instance its own public IPv6 would let the device gateway be reached
+directly, bypassing the NLB.
+
+In the Caddy topology the instance is the only public endpoint, so it gets a
+public IPv6 address of its own (`ipv6_address_count = 1`) alongside its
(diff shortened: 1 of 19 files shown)


StealthyCoder:
SSM manager did not work, this is solved by #265 .

StealthyCoder:
The Google OAuth did not work because the instance could not reach out to Google. Solved by #265

StealthyCoder:
The AWS IAM permissions needed for running init-secrets.sh are not documented. At the minimum you need secretsmanager:CreateSecret.

StealthyCoder:
There was a problem with default instance type maybe being amd64 when you might want to use arm64. This is solved by #265

StealthyCoder:
`name_prefix` needs to match what is in `init-secrets.sh` otherwise the lookup on secrets will fail when executing Terraform .

StealthyCoder:
When using the `examples/with-load-balancer` you need to get the entire directory structure under `contrib/aws/terraform` since we use shared modules. Maybe worth a callout somewhere.

StealthyCoder:
`curl -sI "$HOST/favicon" ` will fail because the endpoint does not support HEAD. The verification command should be: `curl -s "$HOST/favicon" -o /dev/null -w "%{http_code}"`

StealthyCoder:
The `user-add` step assumes local auth, but when using Google OAuth for example this will not work. In addition nowhere yet it seems we explained what auth options are available and how they all work.

StealthyCoder:
The CloudWatch agent is baked into the AMI, but it never starts because the `user_data_replace_on_change = false`.  If you could, just add a oneshot systemd unit.

StealthyCoder:
```
# in provision.sh, after the CloudWatch agent install block:

  log "installing CloudWatch agent activation unit"
  cat > /etc/systemd/system/amazon-cloudwatch-agent-config.service <<'EOF'
  [Unit]
  Description=Configure and start CloudWatch Agent when config is present
  ConditionPathExists=/etc/fioserver/cloudwatch-agent.json
  After=network-online.target amazon-cloudwatch-agent.service
  Wants=network-online.target

  [Service]
  Type=oneshot
  RemainAfterExit=yes
  ExecStart=/opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl \
      -a fetch-config -m ec2 -s -c file:/etc/fioserver/cloudwatch-agent.json

  [Install]
  WantedBy=multi-user.target
  EOF
  systemctl enable amazon-cloudwatch-agent-config.service
  ``` 
  
  for example

StealthyCoder:
Everything works out of the box with the fixes from #265 and I just redid the whole flow. :muscle:

StealthyCoder (approved):

PR #231, opened 2026-08-20
End-to-end tests using aktualizr-lite

detsch:
Also implement ostree/download-urls, required by fiopull


doanac (commented):
I like the general idea. A couple of notes: The comments can become excessive in places thanks to Claude. For example, I'm not sure the fiopull test case needs to explain how fiopull works/differs from the normal path; its sufficient to have two different tests.

The middle commit is a lot to follow. It "contrib" code, so I'm fine with it if its a pain to splite up, but its hard to really follow that commit. I mostly skipped to the end and looked at the test_e2e_aklite_update_flow.py file.

detsch:
> A couple of notes: The comments can become excessive in places thanks to Claude. For example, I'm not sure the fiopull test case needs to explain how fiopull works/differs from the normal path; its sufficient to have two different tests.

Comments adjusted.
 
> The middle commit is a lot to follow. It "contrib" code, so I'm fine with it if its a pain to splite up, but its hard to really follow that commit. I mostly skipped to the end and looked at the test_e2e_aklite_update_flow.py file.

Previous middle commit is now 2 commits: one making the fixture generic, and other adding the test and required files.

mike-sul:
This commit message https://github.com/foundriesio/update-server/pull/231/changes/0f77b31a15c35fc28662f0e4087dbe956a55d760 is a bit misleading as it states that the "download-urls" is needed only for the "fioupull" case while it is a default case for pulling through libostree. Also, we need to make sure this works for both cases, the default (libostree pull) and the fiopull one.

mike-sul:
@detsch 
> Two things worth noting: the client image needs glib-networking because libostree 2025.1 pulls over libsoup3/GnuTLS

I think it is build time configurable, that is why we explicitly specify `--without-soup` option (can be also `--with-curl`) when building libostree for the dev container https://github.com/foundriesio/aktualizr-lite/blob/c44e6dc0d764e73852216994211651a5e1510a2a/docker/Dockerfile#L102.
In Yocto it is done in some clever way in meta-oe, we used to specify it explicitly in  meta-updater too https://github.com/uptane/meta-updater/commit/9bca3262df0c162d210ae036bb20899da3685f5d#diff-ac5a0637ddb46d2bb381d4bda12e3438a3f8301ec342d58046c15a4303aa8007R9.

So, I suggest to build libostree and ostree with curl support instead of soap in the e2e test container too, so we test the same path as we have on a real device.

detsch:
> This commit message [0f77b31](https://github.com/foundriesio/update-server/commit/0f77b31a15c35fc28662f0e4087dbe956a55d760) is a bit misleading as it states that the "download-urls" is needed only for the "fioupull" case while it is a default case for pulling through libostree. Also, we need to make sure this works for both cases, the default (libostree pull) and the fiopull one.

Commit message adjusted.

detsch:
> @detsch
> 
> > Two things worth noting: the client image needs glib-networking because libostree 2025.1 pulls over libsoup3/GnuTLS
> 
> I think it is build time configurable, that is why we explicitly specify `--without-soup` option (can be also `--with-curl`) when building libostree for the dev container https://github.com/foundriesio/aktualizr-lite/blob/c44e6dc0d764e73852216994211651a5e1510a2a/docker/Dockerfile#L102. In Yocto it is done in some clever way in meta-oe, we used to specify it explicitly in meta-updater too [uptane/meta-updater@9bca326#diff-ac5a0637ddb46d2bb381d4bda12e3438a3f8301ec342d58046c15a4303aa8007R9](https://github.com/uptane/meta-updater/commit/9bca3262df0c162d210ae036bb20899da3685f5d#diff-ac5a0637ddb46d2bb381d4bda12e3438a3f8301ec342d58046c15a4303aa8007R9).
> 
> So, I suggest to build libostree and ostree with curl support instead of soap in the e2e test container too, so we test the same path as we have on a real device.

You are right, and that's how we build it already (since this environment extends https://github.com/foundriesio/aktualizr-lite/blob/master/docker/Dockerfile), so it is just a matter of not adding unnecessary packages here. Fixed.

mike-sul:
@detsch I don't see that the e2e test sets `compose_apps_proxy` param in the device .toml conf. If it is not set then aklite pulls/updates apps directly from a registry while the primary use-case is pulling apps from the update-server. I think we should test it first of all.  @doanac FYI.

detsch:
> @detsch I don't see that the e2e test sets `compose_apps_proxy` param in the device .toml conf. If it is not set then aklite pulls/updates apps directly from a registry while the primary use-case is pulling apps from the update-server. I think we should test it first of all. @doanac FYI.

@mike-sul  It is set here: https://github.com/detsch/update-server/blob/bb592ab2c64dc4c31b07ecec621c6996f8efb826/contrib/e2e/conftest.py#L57

Should we test both cases? Or just keep this one?

mike-sul:
> @mike-sul It is set here: https://github.com/detsch/update-server/blob/bb592ab2c64dc4c31b07ecec621c6996f8efb826/contrib/e2e/conftest.py#L57
> 
> Should we test both cases? Or just keep this one?

ok, I suppose just one case (fetching apps from the fio update-server) at this stage.

mike-sul (approved):

PR #232, opened 2026-08-20
ui: Include tip for DEVICE_API endpoint

doanac:
fio-device-register needs the user to set `--device-api`. This isn't totally obvious for a user. Showing this value in the settings is a way to cheat and likely get the user's attention: they'll need to go to the place to generate their API token and will probably see it.

### DIFF
--- a/server/ui/web/handlers_settings.go
+++ b/server/ui/web/handlers_settings.go
@@ -19,14 +19,16 @@ func (h handlers) settings(c echo.Context) error {
 
 	ctx := struct {
 		baseCtx
-		Tokens     []users.Token
-		ScopesList []string
-		LocalAuth  bool
+		Tokens       []users.Token
+		ScopesList   []string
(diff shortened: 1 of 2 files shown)


doanac:
image

doanac:
@mike-scott - a bit of a usability question. I hit this very problem today and was trying to think of a way to better expose this information when we are no longer able to provide defaults for it in our LmP

mike-scott:
@doanac ah, this fills a nice gap.

mike-scott (approved):
LGTM

mike-scott:
Question: Not sure if this can work right through a container (like it might show the internal port that it knows about, but not the mapped port externally)

doanac:
> Question: Not sure if this can work right through a container (like it might show the internal port that it knows about, but not the mapped port externally)

It works in a container as well. We've used this pattern a couple places in the project

PR #236, opened 2026-08-25
ui: fail/panic when UI assets are Git-LFS pointer stubs

mike-scott:
The Pico CSS asset lives in Git-LFS and is baked into the binary with go:embed. A clone made without git-lfs embeds the pointer stub instead, and the UI renders unstyled with no indication why. Per review feedback, this now fails loudly instead of warning.

-  Refuse to start with a stubbed asset and fail the container image build early.
- Add the git-lfs setting to contrib-test, merge, and unit-test which would now fail/panic.

### DIFF
--- a/contrib/server/Dockerfile
+++ b/contrib/server/Dockerfile
@@ -11,6 +11,11 @@ COPY go.* ./
 RUN go mod download
 
 COPY . .
+# go:embed silently bakes Git-LFS pointer stubs into the binary
+RUN if grep -q "git-lfs.github.com/spec" server/ui/web/templates/pico_min_211.css; then \
+        echo "ERROR: LFS assets are pointer stubs - the web UI would be unstyled. Run 'git lfs pull' and rebuild." >&2; \
+        exit 1; \
+    fi
 RUN go build \
     --ldflags '-linkmode external -extldflags "-static"' \
     -o fioserver \
--- a/.github/workflows/contrib-test.yml
+++ b/.github/workflows/contrib-test.yml
@@ -18,6 +18,9 @@ jobs:
           go-version: '~1.24'
       - name: Clone the code
         uses: actions/checkout@v4
+        with:
+          # the server binary and its container build fail on LFS pointer stubs
+          lfs: true
 
       - name: Build and test devshell
         run: |
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -20,6 +20,9 @@ jobs:
 
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          # e2e runs fioserver, which panics on LFS pointer stubs
+          lfs: true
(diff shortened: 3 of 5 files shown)


contrib/server/Dockerfile
```diff
@@ -11,6 +11,10 @@ COPY go.* ./
 RUN go mod download
 
 COPY . .
+# go:embed silently bakes Git-LFS pointer stubs into the binary
+RUN if grep -q "git-lfs.github.com/spec" server/ui/web/templates/pico_min_211.css; then \
+        echo "WARNING: LFS assets are pointer stubs - the web UI will be unstyled. Run 'git lfs pull' and rebuild."; \
```
doanac: I didn't know you could easily detect this. Exit with an error; nobody is going to notice this buried inside all the output from docker-build.

server/ui/server.go
```diff
@@ -28,6 +30,17 @@ type daemon interface {
...
+	// go:embed bakes in whatever bytes the checkout had; a clone without
+	// git-lfs leaves pointer stubs and an unstyled UI
+	if entries, err := templates.Assets.ReadDir("."); err == nil {
+		for _, e := range entries {
+			if data, err := templates.Assets.ReadFile(e.Name()); err == nil &&
+				bytes.HasPrefix(data, []byte("version https://git-lfs.github.com/spec")) {
+				log.Warn("UI asset is a Git-LFS pointer stub; the UI will be unstyled. Run 'git lfs pull' and rebuild", "asset", e.Name())
+			}
+		}
+	}
```
doanac: I think you could move this logic to https://github.com/foundriesio/update-server/blob/main/server/ui/web/templates/templates.go and call `panic`.

doanac (commented):
nice idea - lets be even more aggressive and not warn but fail

contrib/server/Dockerfile
mike-scott: I originally had it as an error, but then I was worried most folks will use a static binary from the release:
https://github.com/foundriesio/update-server/releases/download/v0.9.4/fioserver-linux-amd64
And wasn't sure if that had the same issue (it shouldn't now that I think about it)

Then I was wondering if they were using source: they might not want to install git-lfs for some reason ...
So I decided on the warning.  I can flip it back.

mike-scott: Re-worked to an error.

server/ui/server.go
mike-scott: Moved the panic to `server/ui/web/templates/templates.go` in `init()`

mike-scott:
- Rebased on `main`
- Move the panic to `server/ui/web/templates/templates.go` in `init()`
- Updated the CI workflows to include git-lfs settings (so they don't panic / fail)

doanac (approved):

PR #239, opened 2026-08-25
api: Fix flaky rollout and update-tail tests

mike-scott:
`TestApiRolloutDaemon` and `TestApiUpdateTail` fail intermittently in CI.
Both reproduce on main — the rollout test by timing, the tail test under
`go test -race`.

* **api: Deflake the rollout daemon test** — the test slept 60ms and
  asserted the daemon had committed the rollouts; a loaded runner can
  delay the daemon past that. Poll for the committed state instead.
* **api: Make the keepalive stream interval atomic** — the tail test
  overrides the interval while live stream handlers are reading it; store
  it atomically (the one production-code change).
* **api: Fix data races in the update tail test** — the test read
  `httptest.ResponseRecorder` fields while the SSE handler goroutine was
  still writing, and used fixed sleeps to wait for stream output. Guard
  the recorder with a mutex, poll for expected bodies, and join the
  handler goroutines instead of sleeping.

Both tests now pass 20 consecutive runs under `-race` (previously failing
within 5), and the full `server` and `storage` suites are green under
`-race`.

### DIFF
--- a/server/ui/api/handlers_rollouts.go
+++ b/server/ui/api/handlers_rollouts.go
@@ -15,6 +15,7 @@ import (
 	"regexp"
 	"slices"
 	"strconv"
+	"sync/atomic"
 	"time"
 
 	"github.com/labstack/echo/v4"
@@ -389,8 +390,20 @@ func filterUpdateLogs(uuids []string, reader iter.Seq2[string, error]) iter.Seq2
 
 var (
 	keepaliveResponseText     = ": idle\n\n"
-	keepaliveResponseInterval = 30 * time.Second
+	keepaliveResponseInterval = newAtomicDuration(30 * time.Second)
 )
+
+// atomicDuration is atomic so tests can override the value while stream handlers are reading it.
+type atomicDuration struct{ atomic.Int64 }
+
+func (d *atomicDuration) Store(v time.Duration) { d.Int64.Store(int64(v)) }
+func (d *atomicDuration) Load() time.Duration   { return time.Duration(d.Int64.Load()) }
+
+func newAtomicDuration(d time.Duration) *atomicDuration {
+	v := new(atomicDuration)
+	v.Store(d)
+	return v
+}
 
 func keepaliveReader(reader iter.Seq2[string, error]) iter.Seq2[string, error] {
 	// An HTTP client will disconnect after an idle time while server does not write annything (usually 5 minutes).
@@ -427,7 +440,7 @@ func keepaliveReader(reader iter.Seq2[string, error]) iter.Seq2[string, error] {
 					// Caller signals to stop reading.
 					break LOOP
 				}
-			case <-time.After(keepaliveResponseInterval):
+			case <-time.After(keepaliveResponseInterval.Load()):
 				if !yield(keepaliveResponseText, nil) {
 					break LOOP
 				}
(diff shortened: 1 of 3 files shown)


server/ui/api/handlers_test.go
```diff
@@ -954,18 +954,18 @@ func TestApiRolloutDaemon(t *testing.T) {
...
-	data = tc.GET("/updates/tag2/update2/rollouts/roll2", 200)
-	assert.Equal(t, `{"uuids":["prod1"],"effective-uuids":["prod1"],"committed":true}`, s(data))
-	dev, err = tc.api.DeviceGet("ci1")
-	assert.Nil(t, err)
-	assert.Equal(t, "update1", dev.UpdateName)
-	dev, err = tc.api.DeviceGet("prod1")
-	assert.Nil(t, err)
-	assert.Equal(t, "update2", dev.UpdateName)
+	// After the watchdog daemon processing, rollouts are committed. Poll
+	// rather than sleep: a loaded runner can delay the daemon arbitrarily.
```
doanac: I'd drop that comment - its really context for this review and is covered in the commit message

server/ui/api/handlers_test.go
```diff
@@ -1041,49 +1095,45 @@ data: {"uuid":"test-device-2","correlationId":"uuid-2","target-name":"intel-core
...
 	tc.assertNotDone(done3)
 
 	cancel() // This is where we disconnect, closing all holding handlers.
-	time.Sleep(10 * time.Millisecond)
-	tc.assertDone(done1)
-	tc.assertDone(done2)
-	tc.assertDone(done3)
+	tc.waitDone(done1)
+	tc.waitDone(done2)
+	tc.waitDone(done3)
```
doanac: You can now remove the assertDone method

server/ui/api/handlers_test.go
mike-scott: Dropped the comment addition.

server/ui/api/handlers_test.go
mike-scott: Dropped assertDone()

mike-scott:
- Re-worked the 2nd patch to look better (IMHO)
- Found another race in the server test, so I added a 4th patch (similar use of atomic to protect the err variable).

server/ui/api/handlers_rollouts.go
```diff
@@ -389,9 +390,21 @@ func filterUpdateLogs(uuids []string, reader iter.Seq2[string, error]) iter.Seq2
...
+
+func (d *atomicDuration) Store(v time.Duration) { d.Int64.Store(int64(v)) }
+func (d *atomicDuration) Load() time.Duration   { return time.Duration(d.Int64.Load()) }
+
+func newAtomicDuration(d time.Duration) *atomicDuration {
+	v := new(atomicDuration)
+	v.Store(d)
+	return v
+}
+
```
doanac: is there a reason you changed this from the first version that used an init function? That one was less code and more obvious to me?

mike-scott:
Rebased onto `main`

server/ui/api/handlers_rollouts.go
mike-scott: Heh, I changed it to match closer to the original source and less int64 conversion stuff in the later code.
It was a personal preference.  I can swap it back.

doanac: ah- i sort of see the cleverness now. Either way is fine with me.

doanac (approved):

PR #240, opened 2026-08-25
fix: local run TUF initialization

MiloCasagrande:
* Make sure TUF is initialized for the local dev.
 * Add optional port parameters for starting server.


MiloCasagrande:
This fixes the local dev run script to bring the server up, as it was missing the TUF initialization.

doanac (approved):
I wonder if this should be run every time we call run-local. The other stuff if this script checks before running things. But that's orthogonal to this change.

MiloCasagrande:
I think as a next step I want to make it so that if it finds previous data, it uses that.

PR #241, opened 2026-08-25
doc: few clean bug fixes from playtesting

mike-scott:
- Fix a fioserv typo
- Fix an auth link


mike-scott:
@doanac @kprosise This is the first few "easy" fixes for docs that we can probably land today.

doanac (approved):

PR #248, opened 2026-08-27
fix: skip init steps when data dir already exists

MiloCasagrande:


MiloCasagrande:
Adds a check so that if a local data dir already exists, the run-local script will use it instead of requiring to remove it.

doanac (approved):

Issue #242, opened 2026-08-26
fioconfig: generates a 405 error when it attempts to PATCH /config for device-published files


mike-scott:
## Summary

The update-server has no device-facing write path for config.  `fioconfig` expects `PATCH` `/config` (mTLS gateway, 8443) to merge device-published files into its device config and answer 201.  The server implements config as read-only toward devices.

## Logged from the server

```
{"time":"2026-08-25T16:47:34.868491217-07:00","level":"ERROR","msg":"response","server":"gateway-api","conn_id":"77c7pDhexs","req_id":"ktxmLEYHzI5Q","uri":"/config","method":"PATCH","content-length":"143","status":405,"err":"code=405, message=Method Not Allowed"}
{"time":"2026-08-25T16:47:34.880302813-07:00","level":"INFO","msg":"response","server":"gateway-api","conn_id":"8m8ISb7PoB","req_id":"rPOWaXsxs8h1","uri":"/config","device":"ddc43f38-5d68-4b7b-859a-7fc69db69a30","method":"GET","content-length":"","status":204}
```

## Logged on the device

```
Aug 25 23:47:34 iq-9075-evk fioconfig[1944]: 2026/08/25 23:47:34 INFO Checking in with server
Aug 25 23:47:34 iq-9075-evk fioconfig[1944]: 2026/08/25 23:47:34 INFO Running initialization callback=remote-actions
Aug 25 23:47:34 iq-9075-evk fioconfig[1944]: 2026/08/25 23:47:34 INFO Configured remote actions changed old="" new=diag,reboot
Aug 25 23:47:34 iq-9075-evk fioconfig[1944]: 2026/08/25 23:47:34 ERROR Unable to update url=https://local-server:8443/config status=405 response="{\"message\":\"Method Not Allowed\"}\n"
Aug 25 23:47:34 iq-9075-evk fioconfig[1944]: 2026/08/25 23:47:34 INFO Device has no config defined on server
```

PR #249, opened 2026-08-27
gateway: Add `PATCH /config` support

doanac:
Fixes issue #242


doanac:
@mike-scott - this should fix your issue

mike-scott:
I'll test this in the morning.  Looks great!

mike-scott:
I verified this works on real HW:
- iq-9075-evk running qli-2.0 with Foundries components
- registered to the latest `main` + this patchset
- logged attempts for `PATCH` /config returned 201

mike-scott (approved):
- I didn't see any glaring security holes
- code looks reasonable for handling wireguard and remote-actions (though I'm not that familiar with the code base)

mike-scott:
Claude seems to think we should take a look at the following:
- Two findings look like blockers. The nil-map panic (`handlers_config.go:68`) is reachable from an operator `PUT` with "Files": null and then wedges every subsequent device `PATCH`. And `ConfigCreate` drops the public-key field the real `fioconfig` client sends on first boot — the review verified this against upstream `fioconfig` source — which silently breaks config encryption for devices whose config key differs from their TLS key.
- A cluster of four findings shares one cause: the gateway path re-implements the config write at handler level and misses the UI path's purge, no-change dedup, and locking. Since `fioconfig` `PATCH`es on every check-in, the missing dedup plus never-purged history means device-controlled unbounded journal/blob growth. The last finding suggests how to fix it is: one storage-level `PatchDeviceConfig` primitive both APIs share, which addresses the race, purge, and dedup together.

doanac:
@mike-scott - fixed in force push minus the dedup suggestion;

doanac:
rebased on latest main.

PR #250, opened 2026-08-28
feat: add air live-reload config for local development

MiloCasagrande:
Watches *.go, *.html, and *.css; rebuilds via make fioserver and
restarts the server automatically. Run contrib/run-local.sh once
first to initialise .local-data, then use `air` in place of
manual rebuild+restart cycles.

air must be installed first:

go install github.com/air-verse/air@latest

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/.air.toml
+++ b/.air.toml
@@ -1,0 +1,16 @@
+root = "."
+tmp_dir = "tmp"
+
+[build]
+  # Reuses Makefile's LDFLAGS (embeds git-describe version string).
+  # Run contrib/run-local.sh once first to initialise the data dir.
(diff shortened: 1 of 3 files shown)


MiloCasagrande:
This is mostly for local web dev. Every time you change a CSS file, you need to rebuild the server for changes to be picked up. This automatically "watches" the source and rebuild on the fly, there is a 1-second delay, but it's better than manually restarting.

doanac (approved):
that's cool. i'd never seen it.

PR #253, opened 2026-08-28
api: Deflake the rollout put test

detsch:
TestApiRolloutPut slept a fixed 50ms for the async rollout commit before asserting the committed state, which is racy under load. Poll for the committed state with require.Eventually instead, matching the pattern already used by the rollout daemon and update tail tests.

---

Similar to 84d1590


mike-scott (approved):
LGTM: similar to the fix here:
https://github.com/foundriesio/update-server/blob/main/server/ui/api/handlers_test.go#L1007

server/ui/api/handlers_test.go
```diff
@@ -930,27 +930,22 @@ func TestApiRolloutPut(t *testing.T) {
...
+		dev3, err3 := tc.api.DeviceGet("prod1")
+		dev4, err4 := tc.api.DeviceGet("prod2")
+		dev5, err5 := tc.api.DeviceGet("prod3")
+		return d1 == `{"uuids":["ci1","ci2","ci3"],"effective-uuids":["ci1","ci2"],"committed":true}` &&
+			d2 == `{"uuids":["prod2"],"groups":["grp1"],"effective-uuids":["ci4","prod2","prod3"],"committed":true}` &&
+			err1 == nil && dev1.UpdateName == "update1" &&
+			err2 == nil && dev2.UpdateName == "update1" &&
+			err3 == nil && dev3.UpdateName == "" &&
+			err4 == nil && dev4.UpdateName == "update2" &&
+			err5 == nil && dev5.UpdateName == "update2"
```
doanac: If this test fails - are we going to get useful output for which condition failed? i.e. it looks like if something fails, we'll have to retry with a fmt.Println statement for all these conditions we are testing?

detsch: I've followed the same structure of the previous fixes, but it is possible to have a better output, keeping the individual assertions. Force pushing it now, so we can decide which implementation is better.

doanac (approved):
This seems cleaner - i like it

